PR #7, opened 2022-04-27
feat: allow certificate to be supplied to `serve`

eduardoboucas:

When the `certificatePath` option is supplied to the `serve` method, Edge Bundler will call `deno run` with the `--cert` flag.

Part of https://github.com/netlify/pillar-runtime/issues/323.

### DIFF
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -16,15 +16,17 @@ interface PrepareServerOptions {
   flags: string[]
   formatExportTypeError?: FormatFunction
   formatImportError?: FormatFunction
+  originProtocol?: string
   port: number
 }
 
 const prepareServer = ({
   deno,
   distDirectory,
-  flags,
+  flags: denoFlags,
   formatExportTypeError,
   formatImportError,
+  originProtocol,
   port,
 }: PrepareServerOptions) => {
   const processRef: ProcessRef = {}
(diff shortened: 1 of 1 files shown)


Skn0tt (approved):
looks mostly good to me, left a tiny question

eduardoboucas:
> looks mostly good to me, left a tiny question

I don't see any comments 🤔

Skn0tt:
hm, me neither - GitHub swallowed it? :shrug: give me a sec

src/server/server.ts
```diff
@@ -56,7 +56,9 @@ const prepareServer = ({
       // no-op
     }
 
-    await deno.runInBackground(['run', ...flags, stage2Path, port.toString()], true, processRef)
+    const bootstrapFlags = ['--port', port.toString()]
```
Skn0tt, line 59: looks like we didn't supply the `--port` flag before - assuming this is unrelated to the certificate path, was broken before, and you drive-by-fixed this.

eduardoboucas: We did supply it, but as a positional parameter and not a flag (i.e. `deno run script.js <PORT>` as opposed to `deno run script.js --port <PORT>`). Using flags will allow us to supply additional parameters in the future in a more sustainable way.

You're right that it's not necessarily related to the certificate path, and I should've clarified that. Thanks for calling that out!

eduardoboucas: See https://github.com/netlify/edge-functions-bootstrap/pull/30/files#r860951538.

PR #22, opened 2022-05-04
feat: add trailing slash to regular expressions

eduardoboucas:

This PR changes the regular expressions generated from a `path` declaration so that they contain an optional trailing slash.

So `/foo` will yield `^/foo/?$` instead of `^/foo$`.

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -30,7 +30,7 @@ const generateManifest = ({ bundles = [], declarations = [], functions }: Genera
       return
     }
 
-    const pattern = 'pattern' in declaration ? new RegExp(declaration.pattern) : globToRegExp(declaration.path)
+    const pattern = getRegularExpression(declaration)
     const serializablePattern = pattern.source.replace(/\\\//g, '/')
 
     return {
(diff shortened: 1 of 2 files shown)


src/manifest.ts
```diff
@@ -30,7 +30,7 @@ const generateManifest = ({ bundles = [], declarations = [], functions }: Genera
       return
     }
 
-    const pattern = 'pattern' in declaration ? new RegExp(declaration.pattern) : globToRegExp(declaration.path)
```
charliegroll, line 33: wait what is this `'pattern' in declaration`? is that the same as `declaration?.pattern`? I almost never see `in` used

charliegroll (approved):
not sure how I can verify, but your test updates look kosher to me :shipit:

src/manifest.ts
eduardoboucas: Yes, but since `declaration` is a union type (`DeclarationWithPath` or `DeclarationWithPattern`) and the `pattern` only exists in one of them, the `in` allows us to disambiguate. So if you do:

```ts
if ('pattern' in declaration) {
  // From this point on you know that `declaration` is of type
  // `DeclarationWithPattern` and so does the TypeScript compiler.
}
```

fool:
Do we need to worry about the case where a customer wants different behavior for /path and /path/ ?

mraerino (approved):

Issue #23, opened 2022-05-04
Fix flaky tests


eduardoboucas:
~Currently, our test suite actually downloads the Deno CLI binaries as part of each run. This makes the tests flaky and they frequently fail due to network/IO errors. We should consider mocking the HTTP call and using a local version of the Deno CLI.~

Update: actually this doesn't seem related to downloading the Deno CLI. It's actually failing on a very specific test when trying to create a JavaScript bundle in parallel with an ESZIP bundle (https://github.com/netlify/edge-bundler/runs/6306315074?check_suite_focus=true#step:7:48)

eduardoboucas:
Still not fixed, sadly.

eduardoboucas:
Hopefully #139 fixes this. Let's reopen if we see failures again.

PR #26, opened 2022-05-05
chore: set up Deno in release-please workflow

eduardoboucas:

Runs the CI matrix sequentially, in an attempt to diagnose https://github.com/netlify/edge-bundler/issues/23.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -27,6 +27,11 @@ jobs:
           check-latest: true
           registry-url: 'https://registry.npmjs.org'
         if: ${{ steps.release.outputs.release_created }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release-please.yml
```diff
@@ -27,6 +27,11 @@ jobs:
           check-latest: true
           registry-url: 'https://registry.npmjs.org'
         if: ${{ steps.release.outputs.release_created }}
+      - name: Setup Deno
```
eduardoboucas, line 30: Adding this so that the test suite that runs as part of `npm publish` doesn't have to download the binary.

jackiewmacharia (approved):

PR #31, opened 2022-05-18
feat: add deno --inspect and --inspect-brk flags

jackiewmacharia:

Example: This fixes #5012

**Checklist**

### DIFF
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -75,6 +75,8 @@ interface ServeOptions {
   certificatePath?: string
   debug?: boolean
   distImportMapPath?: string
+  edgeInspect?: boolean | string
+  edgeInspectBrk?: boolean | string
   importMaps?: ImportMapFile[]
   onAfterDownload?: LifecycleHook
   onBeforeDownload?: LifecycleHook
@@ -83,10 +85,13 @@ interface ServeOptions {
   port: number
 }
 
+// eslint-disable-next-line complexity, max-statements
 const serve = async ({
   certificatePath,
   debug,
   distImportMapPath,
+  edgeInspect,
+  edgeInspectBrk,
   formatExportTypeError,
   formatImportError,
   importMaps,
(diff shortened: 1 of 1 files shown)


src/server/server.ts
```diff
@@ -83,10 +85,13 @@ interface ServeOptions {
   port: number
 }
 
+// eslint-disable-next-line complexity, max-statements
 const serve = async ({
   certificatePath,
   debug,
   distImportMapPath,
+  edgeInspect,
```
eduardoboucas, line 93: What happens if we receive both `edgeInspect` and `edgeInspectBrk`?

Rather than receiving two flags that can be either strings or booleans, what do you think of receiving a single variable of an `InspectSettings` type looking something like:

```ts
interface InspectSettings {
  // Whether inspect mode is enabled
  enabled: boolean

  // Whether to pause on breakpoints (i.e. "--brk")
  pause: boolean

  // Host/port override (optional)
  address?: string
}
```

CLI would then compile this object based on the flags it receives and pass it along to Edge Bundler.

jackiewmacharia: It defaults to --inspect-brk if both are received so I've added a conflict check to CLI so both are not passed at the same time. Switched edge-bundler over to a single object as well.

jackiewmacharia:
@eduardoboucas this one's ready

src/server/server.ts
```diff
@@ -122,6 +135,14 @@ const serve = async ({
     flags.push('--quiet')
   }
 
+  if (inspectSettings && inspectSettings.enabled) {
+    if (inspectSettings.pause) {
+      flags.push(inspectSettings.address ? `--inspect-brk${inspectSettings.address}` : '--inspect-brk')
+    } else {
+      flags.push(inspectSettings.address ? `--inspect${inspectSettings.address}` : '--inspect')
```
eduardoboucas, line 142: Shouldn't there be a `=` between the flag and the address? Like `--inspect=1.2.3.4:56`?

jackiewmacharia: The address comes with it when passed from CLI e.g `=127.0.01:3030`

eduardoboucas: But why would we expect an address argument starting with a `=`? It feels like CLI should be stripping it out so that Edge Bundler remains generic.

jackiewmacharia: Stripped out from CLI side now.

PR #42, opened 2022-06-09
fix: add error state to onAfterDownload

jackiewmacharia:

Adds error state to onAfterDownload to allow for icon change in case there is an issue with deno binary download or installation

* Part of https://github.com/netlify/cli/issues/4621
* Related to https://github.com/netlify/cli/pull/4685

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## [1.4.0](https://github.com/netlify/edge-bundler/compare/v1.3.0...v1.4.0) (2022-06-21)
+
+
+### Features
+
+* return generated manifest object ([#48](https://github.com/netlify/edge-bundler/issues/48)) ([a9eadcf](https://github.com/netlify/edge-bundler/commit/a9eadcf74b3baef606afb0ab43535fe4368d3b58))
+
+## [1.3.0](https://github.com/netlify/edge-bundler/compare/v1.2.1...v1.3.0) (2022-06-21)
+
+
+### Features
+
+* remove feature flag eszip ([#50](https://github.com/netlify/edge-bundler/issues/50)) ([da6377b](https://github.com/netlify/edge-bundler/commit/da6377bd6b2ed57215c64f185526e8b881522b1b))
+
 ## [1.2.1](https://github.com/netlify/edge-bundler/compare/v1.2.0...v1.2.1) (2022-06-13)
 
 - updated edge-functions-bootstrap version (https://github.com/netlify/edge-bundler/pull/43)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,6 +39,3 @@ After submitting the pull request, please make sure the Continuous Integration c
 ## Releasing
 
 1. Merge the release PR
-2. Switch to the default branch `git checkout main`
-3. Pull latest changes `git pull`
-4. Publish the package `npm publish`
--- a/deno/bundle.ts
+++ b/deno/bundle.ts
@@ -1,4 +1,4 @@
-import { writeStage2 } from 'https://62a718350450bd00093e1841--edge-bootstrap.netlify.app/bundler/mod.ts'
+import { writeStage2 } from 'https://62ac9c589c16c50008b6ef55--edge-bootstrap.netlify.app/bundler/mod.ts'
 
 const [payload] = Deno.args
 const { basePath, destPath, functions } = JSON.parse(payload)
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@netlify/edge-bundler",
-  "version": "1.2.1",
+  "version": "1.4.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@netlify/edge-bundler",
-      "version": "1.2.1",
+      "version": "1.4.0",
       "license": "MIT",
       "dependencies": {
         "common-path-prefix": "^3.0.0",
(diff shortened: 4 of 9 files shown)


danez (changes requested):
LGTM

I was wondering if it would make more sense to supply the error directly to the function instead of the boolean, but there is only one kind of error and I also don't really see a usecase for that. So I think this is fine the way it is.

eduardoboucas:
> LGTM
> 
> I was wondering if it would make more sense to supply the error directly to the function instead of the boolean, but there is only one kind of error and I also don't really see a usecase for that. So I think this is fine the way it is.

Agreed! It seems to me that the amount of work of adding a `boolean` and an `error` is similar, and the latter gives us more flexibility going forward, so I'd recommend doing it now.

danez (approved):

src/bridge.ts
```diff
@@ -12,13 +12,14 @@ import { getBinaryExtension } from './platform.js'
 const DENO_VERSION_FILE = 'version.txt'
 const DENO_VERSION_RANGE = '^1.20.3'
 
-type LifecycleHook = () => void | Promise<void>
+type OnBeforeDownloadHook = () => void | Promise<void>
+type OnAfterDownloadHook = (hasError: boolean) => void | Promise<void>
```
eduardoboucas, line 16: If we're telling the callback that there is an error, I think we might as well send it.

```suggestion
type OnAfterDownloadHook = (error: error | null) => void | Promise<void>
```

src/bridge.ts
```diff
@@ -65,13 +66,18 @@ class DenoBridge {
     // a malformed semver range. If this does happen, let's throw an error so
     // that the tests catch it.
     if (downloadedVersion === undefined) {
-      throw new Error('Could not read downloaded binary')
+      if (this.onAfterDownload) {
+        this.onAfterDownload(true)
+      }
+      throw new Error(
```
eduardoboucas, line 72: I don't think we should mention CLI here. We're in Edge Bundler, which is also a dependency of Netlify Build, so it could be used in a context that doesn't involve the CLI at all.

eduardoboucas (changes requested):

src/bridge.ts
```diff
@@ -65,13 +66,21 @@ class DenoBridge {
     // a malformed semver range. If this does happen, let's throw an error so
     // that the tests catch it.
     if (downloadedVersion === undefined) {
-      throw new Error('Could not read downloaded binary')
+      const error = new Error(
+        "There was a problem setting up the Edge Functions environment and it's unfortunately not possible to run Edge Functions from this platform. More on supported platforms here: https://deno.land/manual/getting_started/installation.",
```
eduardoboucas, line 70: What do you think about a slightly different spin on the message? My thinking is:

- We should try to point people into self-serving a solution. If they're able to install Deno manually, the problem should go away
- It's possible that the problem comes from an unsupported platform, but we don't know for sure

```suggestion
        "There was a problem setting up the Edge Functions environment. To try a manual installation, visit https://deno.land/manual/getting_started/installation.",
```

We should also try to replace this with a `ntl.fyi` shortlink. I think @minivan is looking into getting us access. If that doesn't happen soon, we can merge as is but we should do a follow up.

eduardoboucas: That's great! I've updated my suggestion accordingly.

src/bridge.ts
```diff
@@ -12,13 +12,14 @@ import { getBinaryExtension } from './platform.js'
 const DENO_VERSION_FILE = 'version.txt'
 const DENO_VERSION_RANGE = '^1.20.3'
 
-type LifecycleHook = () => void | Promise<void>
+type OnBeforeDownloadHook = () => void | Promise<void>
+type OnAfterDownloadHook = (error: Error | null) => void | Promise<void>
```
danez, line 16: Now that we supply the actual error, we can also make `error` optional instead of `Error | null`

```suggestion
type OnAfterDownloadHook = (error?: Error) => void | Promise<void>
```

This ensures better backwards-compatibility, even though this might be not really relevant here as this is only used internally. But also makes the usage a little simpler, see below (or above :)).

src/bridge.ts
```diff
@@ -65,13 +66,21 @@ class DenoBridge {
...
+      }
+
+      throw error
     }
 
     await this.writeVersionFile(downloadedVersion)
 
     if (this.onAfterDownload) {
-      this.onAfterDownload()
+      this.onAfterDownload(null)
```
danez, line 83:
```suggestion
      this.onAfterDownload()
```

Together with the other suggestion.

src/bridge.ts
```diff
@@ -65,13 +66,21 @@ class DenoBridge {
...
     // that the tests catch it.
     if (downloadedVersion === undefined) {
-      throw new Error('Could not read downloaded binary')
+      const error = new Error(
+        'There was a problem setting up the Edge Functions environment. To try a manual installation, visit https://ntl.fyi/install-deno.',
+      )
+
+      if (this.onAfterDownload) {
+        this.onAfterDownload(error)
+      }
```
danez, line 75: [dust] This is highly dusty :) but we could write this as suggested, also in all the other places, which might help to bring down the amount of lines in this file enough to not have to disable eslint `max-lines` on line 213. Though I think it is better to do this in a follow up where we just do this, instead of adding more changes to this PR.

```suggestion
      this.onAfterDownload?.(error)
```

danez (changes requested):

danez (approved):

eduardoboucas (approved):

PR #44, opened 2022-06-13
chore: deploy id automation

Skn0tt:
This PR adds a GH Actions workflow that opens a PR to update the Edge Bootstrap Deploy ID. I plan to either use a `deploy-succeeded` Netlify hook to call this workflow, or to call it from a GH workflow in edge-functions-bootstrap.

cc https://github.com/netlify/edge-bundler/pull/43#issuecomment-1153822651

PR #45, opened 2022-06-13
Update bootrap continue

Skn0tt:
Follow-up to #44, fixing the bugs and making it *actually* work. Proof: #47

### DIFF
--- a/.github/workflows/update-bootstrap-deploy-id.yml
+++ b/.github/workflows/update-bootstrap-deploy-id.yml
@@ -24,7 +24,7 @@ jobs:
         with:
           title: 'chore: update edge-bundler-bootstrap'
           body:
-            'Updates edge-bundler-bootstrap to [`${{ $inputs.bootstrap_deploy_id
+            'Updates edge-bundler-bootstrap to [`${{ inputs.bootstrap_deploy_id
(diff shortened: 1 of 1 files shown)


.github/workflows/update-bootstrap-deploy-id.yml
```diff
@@ -16,18 +16,17 @@ jobs:
 
       - name: Replace Deploy ID
         run:
-          find ./ -type f -exec sed -i 's/https:\/\/.*--edge-bootstrap\.netlify\.app/https://{{
-          $inputs.bootstrap_deploy_id }}--edge-bootstrap.netlify.app/g' {} \;
+          find ./ -type f -exec sed -i 's@https:\/\/.*--edge-bootstrap\.netlify\.app@https://${{
```
eduardoboucas, line 19: How do you feel about replacing this only on specific files? I wonder if this will become annoying if we have a similar URL in a README file or even in another source file where we don't want to do the replacement.

eduardoboucas (approved):

.github/workflows/update-bootstrap-deploy-id.yml
Skn0tt: Thought about this as well. My reasoning was that it's more likely to have these URLs move to different files, than it is that we have a deploy-specific URL in the README.

PR #48, opened 2022-06-14
feat: return generated manifest object

EwanValentine:

- UPDATED: When the edge manifest is generated, this PR returns the generated manifest object, which means we don't have to load the manifest from the generated file in order to debug/inspect the output in other packages where needed.

carlos-capuchin-1024x1024

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -25,20 +25,14 @@ interface BundleOptions {
   onBeforeDownload?: LifecycleHook
 }
 
-interface CommonBundleOptions {
+interface BundleFormatOptions {
   buildID: string
   debug?: boolean
   deno: DenoBridge
   distDirectory: string
   functions: EdgeFunction[]
   featureFlags: Record<string, string>
-}
-
-interface BundleJSOptions {
   importMap: ImportMap
-}
-
-interface BundleESZIPOptions {
   basePath: string
 }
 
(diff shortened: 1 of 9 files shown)


src/bundler.ts
```diff
@@ -24,6 +25,54 @@ interface BundleOptions {
...
+        deno,
+        distDirectory,
+        functions,
+      }),
+    )
+  }
+
+  return bundleOps
+}
+
```
EwanValentine, line 75: I shifted this into its own function because the linter was complaining (rightly so) about the function getting too long. I'm not sure about the name `BundleOpsArgs` though, sounds kinda vague?

eduardoboucas:
> * When the edge manifest is generated, this PR returns the generated manifest object, which means we don't have to load the manifest from the generated file in order to debug/inspect the output in other packages where needed.

I would argue that actually reading the file from its expected location at deploy time would be part of the diagnostic. Any reason you prefer to expose it programmatically and skip that step?

eduardoboucas:
> Hmmm, I figured the contents of the manifest were the important part, rather than the loading of the manifest file itself. So figured it would be cleaner to just return it, rather than reload the contents from the file again. I wasn't 100% sure on that though, so happy to close this and just reload it in the builder

No objections from me! I just wanted to offer an alternative option in case you hadn't considered it, as it could avoid making changes to Edge Bundler.

src/bundler.ts
EwanValentine: I like it! Great shout, I'll have a play around with it now

src/bundler.ts
```diff
@@ -24,6 +25,60 @@ interface BundleOptions {
   onBeforeDownload?: LifecycleHook
 }
 
+interface CommonBundleOptions {
```
eduardoboucas, line 28: Not a huge deal, but do you think we could use the same interface for both formats, even if some properties end up only being used by one of the formats?

EwanValentine: I did wonder if that might be better actually, I'll go with that, for sure

eduardoboucas (approved):
Nice! ✨

PR #50, opened 2022-06-21
feat: remove feature flag eszip

EwanValentine:

Removes the use eszip feature flag.

Closes: [#51](https://github.com/netlify/edge-bundler/issues/51) 

**Checklist**

- [] The status checks are successful (continuous integration). Those can be seen below.

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -7,7 +7,6 @@ import { v4 as uuidv4 } from 'uuid'
 import { DenoBridge, LifecycleHook } from './bridge.js'
 import type { Bundle } from './bundle.js'
 import type { Declaration } from './declaration.js'
-import { FeatureFlags, getFlags } from './feature_flags.js'
 import { findFunctions } from './finder.js'
 import { bundle as bundleESZIP } from './formats/eszip.js'
 import { bundle as bundleJS } from './formats/javascript.js'
@@ -18,7 +17,6 @@ interface BundleOptions {
   cacheDirectory?: string
   debug?: boolean
   distImportMapPath?: string
-  featureFlags?: FeatureFlags
   importMaps?: ImportMapFile[]
   onAfterDownload?: LifecycleHook
   onBeforeDownload?: LifecycleHook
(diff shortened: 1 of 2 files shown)


Skn0tt (dismissed):

EwanValentine:
@Skn0tt yep! Just spotted those, they didn't fail locally weirdly! On it 🙂

src/bundler.ts
```diff
@@ -65,21 +54,16 @@ const bundle = async (
       functions,
       importMap,
     }),
+    bundleESZIP({
```
eduardoboucas, line 57: This means that we'll always generate two bundles (JS and ESZIP), which in the long term we don't want to do. What do you think about changing the logic so that:

- we use `true` as the default value for `edge_functions_produce_eszip`, so that it'll be enabled for CLI deployments
- if the `edge_functions_produce_eszip` flag is enabled, we generate _only_ ESZIP
- if the `edge_functions_produce_eszip` flag is disabled, we generate _only_ JS — this allows us to still flip a switch and go back to JS bundles in buildbot if something goes wrong

eduardoboucas:
(Also make sure to rename the PR to `feat:` before merging, so that release-please works as expected.)

src/bundler.ts
EwanValentine: Ohhhh gotcha, yeah makes sense to me, good shout!

EwanValentine: Fixed: https://github.com/netlify/edge-bundler/pull/50/commits/da27e7af839dc85c4e88ba0d4b65d50728575ecc

src/bundler.ts
```diff
@@ -24,6 +25,35 @@ interface BundleOptions {
...
+    declarations,
+    distDirectory,
+    functions,
+  })
+
+  if (distImportMapPath && importMap) {
+    await importMap.writeToFile(distImportMapPath)
+  }
+}
+
```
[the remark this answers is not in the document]
eduardoboucas: How do you feel about it? I would actually prefer to ignore the rule in this case, because it feels a bit arbitrary to break this bit out into a separate function just to satisfy the linter, and I think it actually makes the code a bit more difficult to follow.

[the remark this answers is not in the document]
EwanValentine: I definitely agree, wasn't sure how we felt about ignoring the linter though, but there didn't really seem to be a particularly clean split anywhere. I'll update to ignore it for this function

eduardoboucas (approved):
Great work!

PR #61, opened 2022-06-28
chore: make wording around ImportMap value precedence more precise

Skn0tt:
This PR updates a comment to make it clear why we're overriding inputs they way we're doing it.


eduardoboucas (commented):
This was intentional, so we can guarantee that people aren't overriding the `netlify:edge` identifier.

Skn0tt:
Ahh, I see! That makes sense. Is there a workaround for testing out a bootstrap version on a prod site?

eduardoboucas:
> Ahh, I see! That makes sense. Is there a workaround for testing out a bootstrap version on a prod site?

I think the best way is to change your code so that you import the types from the bootstrap URL you're testing rather than `netlify:edge`.

src/import_map.ts
```diff
@@ -17,9 +17,9 @@ class ImportMap {
   constructor(input: ImportMapFile[] = []) {
     const inputImports = input.reduce((acc, { imports }) => ({ ...acc, ...imports }), {})
 
-    // `DEFAULT_IMPORTS` must come last because we want our internal imports to
-    // take precedence.
-    this.imports = { ...inputImports, ...DEFAULT_IMPORTS }
+    // `INTERNAL_IMPORTS` must come last,
```
eduardoboucas, line 20: Isn't this saying the same thing? 🧐

Skn0tt: but happy to defer to you

Skn0tt:
> we can guarantee that people aren't overriding the netlify:edge identifier

How are we guaranteeing this for CLI deploys? I guess if `edge-bundler` is running on the user's machine, they could just go to `node_modules` and edit the code?

eduardoboucas (approved):

eduardoboucas:
> > we can guarantee that people aren't overriding the netlify:edge identifier
> 
> How are we guaranteeing this for CLI deploys? I guess if `edge-bundler` is running on the user's machine, they could just go to `node_modules` and edit the code?

I'm not sure why they'd go through all that trouble when all they're getting is a different set of types for the context object. But no, there's nothing stopping people from doing that.

Skn0tt:
Was thinking about security. This means we cannot trust that the edge-bootstrap code is executed for every edge function

PR #66, opened 2022-07-01
feat: refresh types on Deno CLI cache

eduardoboucas:
As part of https://github.com/netlify/pillar-runtime/issues/322, we'd like customers to start importing the types from `https://edge.netlify.com` rather than `netlify:edge`. To ensure that they always have the latest version of that specifier, and not an outdated version cached by Deno CLI, this PR adds a new `ensureLatestTypes` function that will:

1. Hit the `/version.txt` endpoint exposed by https://github.com/netlify/edge-functions-bootstrap/pull/57 to see what is the latest version of the types
2. Check a `types-version.txt` file locally to see whether the cache was already refreshed for that version
3. If not, run `deno cache -r https://edge.netlify.com` to update the specifier in the Deno CLI cache

`ensureLatestTypes` runs on both `bundle` and `serve`, and any errors resulting from the operation will be handled, so that the application will continue to operate normally if something goes wrong with the HTTP call or the cache refresh.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,0 +1,78 @@
+import { promises as fs } from 'fs'
+import { join } from 'path'
+
+import fetch from 'node-fetch'
+
+import type { DenoBridge } from './bridge.js'
+
+const TYPES_URL = 'https://edge.netlify.com'
+
+// eslint-disable-next-line complexity, max-statements
+const ensureLatestTypes = async (deno: DenoBridge, customTypesURL?: string) => {
+  const typesURL = customTypesURL ?? TYPES_URL
+
+  let [localVersion, remoteVersion] = [await getLocalVersion(deno), '']
+
+  try {
+    remoteVersion = await getRemoteVersion(typesURL)
+  } catch (error) {
+    deno.log('Could not check latest version of types:', error)
+
+    return
+  }
+
+  if (localVersion === remoteVersion) {
+    deno.log('Local version of types is up-to-date:', localVersion)
+
+    return
+  }
+
+  deno.log('Local version of types is outdated, updating:', localVersion)
+
+  try {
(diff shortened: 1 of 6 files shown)


src/types.ts
```diff
@@ -0,0 +1,78 @@
+import { promises as fs } from 'fs'
+import { join } from 'path'
+
+import fetch from 'node-fetch'
+
+import type { DenoBridge } from './bridge.js'
+
+const TYPES_URL = 'https://edge.netlify.com'
+
+// eslint-disable-next-line complexity, max-statements
```
Skn0tt, line 10: should we disable complexity and max-statements globally? this is the sort of thing that code reviews are better at catching

test/types.ts
```diff
@@ -0,0 +1,82 @@
...
+  t.is(versionFile, mockVersion)
+
+  mock.restore()
+
+  await fs.rmdir(tmpDir.path, { recursive: true })
+})
+
+test('`ensureLatestTypes` does not update the Deno CLI cache if the local version of types is up-to-date', async (t) => {
+  const mockURL = 'https://edge.netlify'
+  const mockVersion = '987654321'
```
Skn0tt, line 40: that's my favourite version 😄

Skn0tt (approved):
lgtm!

src/types.ts
eduardoboucas: Good call! Done in https://github.com/netlify/edge-bundler/pull/66/commits/bb8aaf54251f32317f0e5de19f69bbad45010bad. We should look into disabling that at in https://github.com/netlify/eslint-config-node/, really.

test/types.ts
eduardoboucas: 😎

PR #76, opened 2022-07-22
feat: accept `basePath` parameter

eduardoboucas:

Adds a new `basePath` property to the `bundle` function. When present, that is used instead of the default "common path" approach.

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -17,6 +17,7 @@ import { writeManifest } from './manifest.js'
 import { ensureLatestTypes } from './types.js'
 
 interface BundleOptions {
+  basePath?: string
   cacheDirectory?: string
   debug?: boolean
   distImportMapPath?: string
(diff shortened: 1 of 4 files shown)


src/bundler.ts
```diff
@@ -155,7 +157,12 @@ const createFinalBundles = async (bundles: Bundle[], distDirectory: string, buil
   await Promise.all(renamingOps)
 }
 
-const getBasePath = (sourceDirectories: string[]) => {
+const getBasePath = (sourceDirectories: string[], inputBasePath?: string) => {
+  // If there's a specific base path supplied, that takes precedence.
+  if (inputBasePath !== undefined) {
```
danez, line 162: Should we do 
```suggestion
  if (inputBasePath != undefined) {
```
in case it is null?

eduardoboucas: The types are saying that it's either `string` or `undefined`, so `null` feels like a case that we shouldn't worry about?

danez: True.

danez (approved):

PR #80, opened 2022-07-28
feat: add support for `denoDir` parameter

eduardoboucas:

Adds a `denoDir` parameter to the `bundle` function. When set, its value is used to set the `DENO_DIR` environment variable.

https://github.com/netlify/pod-compute/issues/161

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -4,7 +4,7 @@ import { join } from 'path'
 import commonPathPrefix from 'common-path-prefix'
 import { v4 as uuidv4 } from 'uuid'
 
-import { DenoBridge, OnAfterDownloadHook, OnBeforeDownloadHook } from './bridge.js'
+import { DenoBridge, DenoOptions, OnAfterDownloadHook, OnBeforeDownloadHook } from './bridge.js'
 import type { Bundle } from './bundle.js'
 import type { Declaration } from './declaration.js'
 import { EdgeFunction } from './edge_function.js'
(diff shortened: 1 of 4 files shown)


src/bundler.ts
```diff
@@ -94,12 +94,18 @@ const bundle = async (
...
+  const options: DenoOptions = {
     debug,
     cacheDirectory,
     onAfterDownload,
     onBeforeDownload,
-  })
+  }
+
+  if (cacheDirectory !== undefined && featureFlags.edge_functions_cache_deno_dir) {
+    options.denoDir = join(cacheDirectory, 'deno_dir')
```
danez, line 105: `cacheDirectory` is set by build to a cachable directory?

eduardoboucas: Yes. See [here](https://github.com/netlify/build/blob/main/packages/build/src/plugins_core/edge_functions/index.js#L44-L47) and [here](https://github.com/netlify/build/blob/main/packages/build/src/plugins_core/edge_functions/index.js#L55).

danez: Could that potentially mean that dependencies which are cached to never get updated? or is the full version always included in the URL?
Kind of unrelated:
Same actually for the cli, if we cache it in buildbot it will never get updated? Also why not install the deno CLI in the build-image directly, to save time on caching and downloading/installing deno? Sorry for these questions just trying to understand where this is all coming from.

eduardoboucas: > Could that potentially mean that dependencies which are cached to never get updated? or is the full version always included in the URL?

Yes, the version should be included in the URL. Having a permanent directory that the Deno CLI can use to cache different things is its normal operation mode. This means not only caching remote dependencies, but also the result of compiling TypeScript to JavaScript.

Currently, we're forcing it to "start from scratch" on every build, which comes with a performance penalty.

> Same actually for the cli, if we cache it in buildbot it will never get updated?

We store the version of the CLI that is cached locally and we compare it against the version that Edge Bundle requires. If there's not a match, we download a fresh version. You can see that logic [here](https://github.com/netlify/edge-bundler/blob/main/src/bridge.ts#L98-L116).

> Also why not install the deno CLI in the build-image directly, to save time and traffic on caching and downloading/installing deno?

See https://github.com/netlify/pillar-workflow/issues/504.

> Sorry for these questions just trying to understand the history.

No problem at all. These are great questions.

danez (approved):
Lgtm

PR #82, opened 2022-08-01
feat: allow setting environment variables and disable extending in DenoServer

danez:
This is a prerequisite for https://github.com/netlify/cli/issues/4614.

The PR allows setting environment variables in `deno.run` and `deno.runInBackground` as well as disabling extending of current runtime environment variables. These two options basically are plain `execa` options, which are forwarded to execa.

I disabled the extending of env variables in the DenoServer so that once the CLI sets the variables it will be only those available in deno. With the exception of `DENO_DIR`. `DENO_DIR` is okay to have locally?

### DIFF
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -31,6 +31,8 @@ interface ProcessRef {
 
 interface RunOptions {
   pipeOutput?: boolean
+  env?: NodeJS.ProcessEnv
+  extendEnv?: boolean
 }
 
 class DenoBridge {
@@ -186,8 +188,8 @@ class DenoBridge {
     return { global: false, path: downloadedPath }
   }
 
-  getEnvironmentVariables() {
-    const env: Record<string, string> = {}
+  getEnvironmentVariables(inputEnv: NodeJS.ProcessEnv = {}) {
+    const env: NodeJS.ProcessEnv = { ...inputEnv }
 
     if (this.denoDir !== undefined) {
       env.DENO_DIR = this.denoDir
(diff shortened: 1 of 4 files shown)


test/main.ts
```diff
@@ -17,7 +17,7 @@ const archiver = require('archiver')
 
 test('Downloads the Deno CLI on demand and caches it for subsequent calls', async (t) => {
   const latestVersion = '1.20.3'
-  const mockBinaryOutput = `#!/usr/bin/env node\n\nconsole.log("deno ${latestVersion}")`
+  const mockBinaryOutput = `#!/usr/bin/env sh\n\necho "deno ${latestVersion}"`
```
danez, line 20: I had to change on test, because the mock for the deno cli was using node, but node couldn't be found anymore, because PATH is not set

danez:
Okay I finally managed to add some tests. This is ready to review.

src/bridge.ts
```diff
@@ -31,6 +31,8 @@ interface ProcessRef {
 
 interface RunOptions {
   pipeOutput?: boolean
+  env?: NodeJS.ProcessEnv
+  extendEnv?: boolean
```
Skn0tt, line 35: In what case would we set `extendEnv` to `true`? Or is this for backwards compatibility?

danez: The default value is `true` atm. I was hardcoding this at first, but then wondered if `runInBackground` might be used in other scenarios in the future. Do you think I should hardcode this? 
And because `run` does the same, besides the background part I also added the options there.

test/bridge.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  if [ "$1" = "test"  ]
+  then
+    env
+  else
+    echo "deno @@@latestVersion@@@"
+  fi`,
+  )
+
+  // The environment sets some varaibles so let us see what they are and remove them from the result
```
eduardoboucas, line 61:
```suggestion
  // The environment sets some variables so let us see what they are and remove them from the result
```

eduardoboucas (approved):

src/server/server.ts
```diff
@@ -60,7 +60,11 @@ const prepareServer = ({
 
     const bootstrapFlags = ['--port', port.toString()]
 
-    await deno.runInBackground(['run', ...denoFlags, stage2Path, ...bootstrapFlags], true, processRef)
+    await deno.runInBackground(['run', ...denoFlags, stage2Path, ...bootstrapFlags], processRef, {
```
eduardoboucas, line 63: I wonder if it's useful to add something like this, for clarity?

```suggestion
    // We set `extendEnv: false` to avoid polluting the edge function context
    // with variables from the user's system, since those will not be available
    // in the production environment.
    await deno.runInBackground(['run', ...denoFlags, stage2Path, ...bootstrapFlags], processRef, {
```

eduardoboucas (approved):

eduardoboucas (approved):

PR #83, opened 2022-08-02
feat: retry mechanism for Deno cli run

EwanValentine:

Due to the occasional network blip, we see failures on build when modules are fetched from Deno. Given the intermittent nature of these issues, it seems like an ephemeral issue which could be solved using a retry mechanism. I've added the most common/popular retry library I could find and have added that around eszip and js bundling.   

This fixes https://github.com/netlify/pod-compute/issues/165

I ran this locally and confirmed that both eszip and JS generated their respective bundles. 

**Checklist**

image-asset

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "1.8.0",
       "license": "MIT",
       "dependencies": {
+        "async-retry": "^1.3.3",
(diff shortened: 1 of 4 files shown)


EwanValentine:
Closing, as we're moving this else-where.

PR #84, opened 2022-08-03
feat: update bootstrap url

khendrikse:

This has to do with https://github.com/netlify/pod-compute/issues/159
**Checklist**

cute animal


Skn0tt (approved):

Skn0tt:
note to self: instead of checking both changed URLs manually with curl, I could've relied on the fact that tests are passing

PR #85, opened 2022-08-03
feat: add support for system logger

eduardoboucas:

With https://github.com/netlify/build/issues/3220 in the pipeline, we'll be able to capture debug data into our system logs without polluting customer builds logs. This PR takes advantage of that and introduces an optional `logFunction` to both `bundle` and `serve`, which is then used instead of `console.log` to print logs.

When this is in place, Netlify Build will be able to pass along the `systemLog()` utility, which was introduced in https://github.com/netlify/build/pull/4406.

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -1,16 +1,25 @@
 type LogFunction = (...args: unknown[]) => void
 
-interface Logger {
-  debug: LogFunction
-  error: LogFunction
-  log: LogFunction
+const noopLogger: LogFunction = () => {
+  // no-op
 }
 
-const getLogger = (parentLogger: LogFunction = console.log): Logger => ({
-  debug: (...args: unknown[]) => parentLogger({ level: 'debug' }, ...args),
-  error: (...args: unknown[]) => parentLogger({ level: 'error' }, ...args),
-  log: (...args: unknown[]) => parentLogger({ level: 'info' }, ...args),
-})
+interface Logger {
+  system: LogFunction
+  user: LogFunction
+}
+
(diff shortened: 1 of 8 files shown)


danez (approved):
Did we consider using a library for the logger? Like winston or pino? Might be an overkill, but both have support for multiple transport, so different log levels can be routed to different transports.

Anyway, might be nice in the future to pass this down to zisi too, so we can log more debug messages there too. :)

eduardoboucas:
Sorry @danez, I rejigged things quite a bit, because I was holding this wrong. I think more than debug levels (which indeed we'd probably be better off getting from a logging library), we need the ability to differentiate user-facing errors from system errors.

Can you have another look at this PR, please? I'll be happy to answer any questions that you may still have after reviewing.

src/logger.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const getLogger = (systemLogger?: LogFunction, debug = false): Logger => {
+  // If there is a system logger configured, we'll use that. If there isn't,
+  // we'll pipe system logs to stdout if `debug` is enabled and swallow them
+  // otherwise.
+  const system = systemLogger ?? (debug ? console.log : noopLogger)
+
+  return {
+    system,
+    user: console.log,
```
danez, line 20: We do not use that anywhere yet?

eduardoboucas: We don't. I thought it'd be nice to have it here for when we do need it, but I can also remove it if you prefer.

danez: No all fine just wondering.

danez (approved):

PR #93, opened 2022-08-09
fix: ensure PATH is always set

danez:
PR #82 broke the detection of global installed binaries. The detection of the binary uses all environment variables and finds the global binary. When we then run the edge function with `deno ...` deno is not found because PATH is not there anymore.

I don't like this solution, but I like the alternative even less. The only alternative i could come up is using `node-which` package to find the absolute path to the global binary. But not sure if we really want to add that overhead?
What do you think? This means that edge functions in dev will have PATH available.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
     "node-fetch": "^3.1.1",
     "node-stream-zip": "^1.15.0",
     "p-wait-for": "^4.1.0",
+    "path-key": "^4.0.0",
     "semver": "^7.3.5",
     "tmp-promise": "^3.0.3",
     "uuid": "^8.3.2"
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -3,6 +3,7 @@ import path from 'path'
 import process from 'process'
 
 import { execa, ExecaChildProcess, Options } from 'execa'
+import pathKey from 'path-key'
 import semver from 'semver'
 
 import { download } from './downloader.js'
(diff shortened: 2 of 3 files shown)


eduardoboucas (dismissed):

danez:
Okay after thinking a lot about it, I guess this is the better solution to correctly resolve the global binary and not rely on PATH being present.

src/bridge.ts
```diff
@@ -199,6 +200,9 @@ class DenoBridge {
       env.DENO_DIR = this.denoDir
     }
 
+    // Ensure PATH is always set as otherwise we are not able to find the global deno binary
+    env[pathKey()] = inputEnv[pathKey({ env: inputEnv })] || process.env[pathKey()]
```
eduardoboucas, line 204: Doesn't `pathKey` always need the environment variables passed in?

If I look at [this](https://github.com/sindresorhus/path-key/blob/main/index.js#L11) correctly, and if the variable happens to be `PATH` on Windows, you're then setting it as `Path` in the final env object?

Or am I misunderstanding? 🤔

package.json
```diff
@@ -87,6 +87,7 @@
     "node-fetch": "^3.1.1",
     "node-stream-zip": "^1.15.0",
     "p-wait-for": "^4.1.0",
+    "path-key": "^4.0.0",
```
eduardoboucas, line 90: The diff in `package-lock.json` seems pretty large. Are you sure that's all due to the addition of `path-key`?

src/bridge.ts
danez: It sets path to whatever the `PATH` casing is in `process.env`.
So if someone passes in PATH, ti will be converted to `Path` on windows, or whatever the current casing is.

package.json
danez: Ah that might have been the forth an back between different dependencies. Will fix.

eduardoboucas (approved):

PR #100, opened 2022-08-16
fix: implement retries for Deno CLI download and additional logging

jackiewmacharia:

This PR adds additional system logging for Edge Functions bundling and adds retries to Deno CLI download.

If retries are triggered for download, logger.system will have a record of retry attempts similar to the below screenshot:

Screenshot 2022-08-16 at 17 14 54

Relates to https://github.com/netlify/pillar-support/issues/22

--------

Example: This fixes #5012

**Checklist**

cute animal

### DIFF
--- a/src/downloader.ts
+++ b/src/downloader.ts
@@ -3,6 +3,7 @@ import path from 'path'
 
 import fetch from 'node-fetch'
 import StreamZip from 'node-stream-zip'
+import pRetry from 'p-retry'
 import semver from 'semver'
 
 import { getBinaryExtension, getPlatformTarget } from './platform.js'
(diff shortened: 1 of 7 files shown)


src/downloader.ts
```diff
@@ -31,17 +33,26 @@ const download = async (targetDirectory: string, versionRange: string) => {
...
-  const url = getReleaseURL(version)
-  const res = await fetch(url)
-
-  if (res.body === null) {
-    throw new Error('Could not download Deno')
-  }
-
-  return res.body
-}
+const downloadVersion = async (versionRange: string, logger: Logger) =>
```
eduardoboucas, line 36: Totally not a blocker, but personally I think it'd be nicer to keep the retry logic separate from the download logic. You could keep `downloadVersion` as is, and then have `downloadVersionWithRetry` which just something like `pRetry(downloadVersion, { ... })`.

Up to you though!

eduardoboucas (dismissed):

src/downloader.ts
jackiewmacharia: Changed it, which also reset the review.

eduardoboucas (dismissed):
Nice!

eduardoboucas (approved):
✨

PR #102, opened 2022-08-22
feat: inline stage 2 ESZIP bundler

eduardoboucas:

When bundling an edge function with the ESZIP format, Edge Bundler is fetching some logic from the edge bootstrap in order to generate the stage 2 file. In reality, this logic has very little to do with the bootstrap layer itself and is actually concerned with bundling, so it makes sense to move it to Edge Bundler.

By moving it, we're also reducing the number of network calls we need to make at build time, which are causing occasional errors.

### DIFF
--- a/deno/lib/common.ts
+++ b/deno/lib/common.ts
@@ -1,0 +1,47 @@
+import { load } from "https://deno.land/x/eszip@v0.18.0/loader.ts";
+import { LoadResponse } from "https://deno.land/x/eszip@v0.18.0/mod.ts";
+import * as path from "https://deno.land/std@0.127.0/path/mod.ts";
+import { retryAsync } from "https://deno.land/x/retry@v2.0.0/mod.ts";
+
+const inlineModule = (
+  specifier: string,
+  content: string,
+): LoadResponse => {
+  return {
+    content,
+    headers: {
+      "content-type": "application/typescript",
+    },
+    kind: "module",
+    specifier,
+  };
+};
+
+const loadFromVirtualRoot = async (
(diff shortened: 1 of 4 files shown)


deno/lib/common.ts
```diff
@@ -0,0 +1,47 @@
+import { load } from "https://deno.land/x/eszip@v0.18.0/loader.ts";
```
eduardoboucas, line 1: This file was copied untouched from the edge bootstrap repo.

Some of the functions here are used by the edge bootstrap when generating the stage 1 ESZIP. We can live with the duplicated code for now, and later on we can change edge bootstrap so that the files are loaded from this repo instead. Conceptually, it makes more sense for the bootstrap to load bundling logic from the bundler repo than the other way around.

deno/lib/consts.ts
```diff
@@ -0,0 +1,4 @@
+export const PUBLIC_SPECIFIER = "netlify:edge";
```
eduardoboucas, line 1: This file was copied untouched from the edge bootstrap repo.

deno/lib/stage2.ts
```diff
@@ -0,0 +1,81 @@
+import { build, LoadResponse } from 'https://deno.land/x/eszip@v0.18.0/mod.ts'
```
eduardoboucas, line 1: This file was copied untouched from the edge bootstrap repo.

EwanValentine (approved):
Given this was mostly copied from another repo, approving as the change of location for that code makes sense

PR #104, opened 2022-08-24
fix: fix deno download retry logic

danez:
This fixes the retry logic that was introduced in #100.

The old logic would only retry if the fetch throws or the body would be empty, but not if the stream of the request body emits an error, which I think might be happening.

This PR now does two things:

- Make the retry enclose the whole download method including fs-operations like unzipping, etc. This ensures now that if the stream emits an error we also retry in this case. If you think we should not add all fs-operations in the retry logic, then we can do that, but the diff gets bigger as I have to split up a bunch of methods.

- If the fetch request does not return a 200 status-code we instantly error. Right now we do read the body and try to create the zip which fails, because the body most likely is not a valid zip.

I also added tests, which took the longest here :)

Not sure if the tests work on windows, but please start reviewing, even if windows is failing.


src/downloader.ts
```diff
@@ -9,57 +10,62 @@ import semver from 'semver'
...
+      data.on('error', reject)
+      file.on('finish', resolve)
+    })
+
+    await extractBinaryFromZip(zipPath, binaryPath, binaryName)
+
+    return binaryPath
+  } finally {
+    // Try closing and deleting the zip file in an case, error or not
+    await promisify(file.close.bind(file))()
```
eduardoboucas, line 40: Is there a chance that this not fire and we'll await indefinitely?

src/downloader.ts
```diff
@@ -9,57 +10,62 @@ import semver from 'semver'
...
+      data.pipe(file)
+      data.on('error', reject)
+      file.on('finish', resolve)
+    })
+
+    await extractBinaryFromZip(zipPath, binaryPath, binaryName)
+
+    return binaryPath
+  } finally {
+    // Try closing and deleting the zip file in an case, error or not
```
eduardoboucas, line 39:
```suggestion
    // Try closing and deleting the zip file in any case, error or not
```

src/downloader.ts
```diff
@@ -9,57 +10,62 @@ import semver from 'semver'
...
 }
 
 const downloadVersion = async (versionRange: string) => {
   const version = await getLatestVersionForRange(versionRange)
   const url = getReleaseURL(version)
   const res = await fetch(url)
 
-  if (res.body === null) {
-    throw new Error('Could not download Deno')
+  if (res.body === null || res.status !== 200) {
```
eduardoboucas, line 55: This feels a bit too restrictive. Should it be something like `res.status < 200 || res.status > 299`?

src/downloader.ts
danez: Well I hope not. And that would be very unexpected behavior. If an error happens, then the `callback(err)` would be called -> reject.

src/downloader.ts
```diff
@@ -9,57 +10,63 @@ import semver from 'semver'
...
+
+    return binaryPath
+  } finally {
+    // Try closing and deleting the zip file in any case, error or not
+    await promisify(file.close.bind(file))()
+
+    try {
+      await fs.unlink(zipPath)
+    } catch {
+      // no-op
```
eduardoboucas, line 45: Not a blocker, but should we use the system logger to print this error?

test/downloader.ts
```diff
@@ -0,0 +1,151 @@
...
+import { execa } from 'execa'
+import nock from 'nock'
+import tmp from 'tmp-promise'
+
+import { download } from '../src/downloader.js'
+import { getLogger } from '../src/logger.js'
+import { getPlatformTarget } from '../src/platform.js'
+
+const testLogger = getLogger(() => {
+  // console.log(...args)
```
eduardoboucas, line 16: Do you want to remove this?

eduardoboucas (dismissed):

src/downloader.ts
danez: Not sure for this error, because if there was an error with the download, let's say deno returned a 500, then this will add additional noise because the zip wasn't even created.
Also cannot think of a case where this would be a relevant message, as this is simply a cleanup.

What do you think?

test/downloader.ts
danez:
```suggestion
```

src/downloader.ts
eduardoboucas: Good point! Let's leave it.

eduardoboucas (approved):

PR #116, opened 2022-09-14
feat: add tests for `serve` functionality

Skn0tt:
- chore: add test for `serve`
- chore: remove unneeded `await`

Closes https://github.com/netlify/pod-compute/issues/103.


EwanValentine (approved):

EwanValentine:
Looks good, looks like the Windows tests are failing in the most vague way possible, though

test/serving.ts
```diff
@@ -0,0 +1,44 @@
...
+import fetch from 'node-fetch'
+
+import { serve } from '../src/index.js'
+
+const url = new URL(import.meta.url)
+const dirname = fileURLToPath(url)
+const fixturesDir = resolve(dirname, '..', 'fixtures')
+
+test.serial('bundler serving functionality', async (t) => {
+  const PORT_THATS_HOPEFULLY_FREE = 5832
```
eduardoboucas, line 14: Can we call this `port`?

Also, should we use something like https://www.npmjs.com/package/get-port rather than hardcoding a port and hoping that it's free?

eduardoboucas (changes requested):

test/serving.ts
Skn0tt: good call, yeah :D done in https://github.com/netlify/edge-bundler/pull/116/commits/709cc70610a8223550d6f2e915d23e5eecfcbb7e

Skn0tt:
> Looks good, looks like the Windows tests are failing in the most vague way possible, though

I think that's unrelated to this PR, see https://netlify.slack.com/archives/C0359548J07/p1663067013856019

eduardoboucas (dismissed):
Nice one! You might need to move the test file to `test/node/` if https://github.com/netlify/edge-bundler/pull/122 is merged first.

eduardoboucas (approved):
LGTM. I think you should change this to `feat:` now that you're changing runtime code, though.

PR #117, opened 2022-09-14
fix: create `shared` directory

eduardoboucas:
Follow up on #113. This PR organises the codebase into 3 directories:

- `node/`: Contains the Node source code, which will be published to npm as `@netlify/edge-bundler`
- `deno/`: Contains the Deno source code, to be included as part of the npm module
- `shared/`: Contains runtime-agnostic code, to be shared between the Node and Deno programs


karagulamos (approved):

mraerino (approved):
is this a breaking change because the import paths for deno change?

PR #118, opened 2022-09-14
fix: fix `getPackageVersion`

eduardoboucas:
With #117, the internal directory structure changed and therefore the relative location of `package.json` has changed. This PR changes `getPackageVersion` so that it can find `package.json` regardless of the relative path.

### DIFF
--- a/node/package_json.ts
+++ b/node/package_json.ts
@@ -1,8 +1,22 @@
-import { createRequire } from 'module'
+import { readFileSync } from 'fs'
 
-const require = createRequire(import.meta.url)
-const pkgJson = require('../package.json')
+import { findUpSync } from 'find-up'
 
-const getPackageVersion = (): string => pkgJson.version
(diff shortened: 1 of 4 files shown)


node/package_json.ts
```diff
@@ -1,8 +1,22 @@
-import { createRequire } from 'module'
+import { readFileSync } from 'fs'
 
-const require = createRequire(import.meta.url)
-const pkgJson = require('../package.json')
+import { findUpSync } from 'find-up'
 
-const getPackageVersion = (): string => pkgJson.version
+const getPackageVersion = () => {
```
eduardoboucas, line 5: This could use a refactor to make it async and return something better than an empty string when something goes wrong, but that involves changing Netlify Build and I'd like to avoid doing that today if possible. I'll revisit this some other time.

karagulamos (approved):

PR #122, opened 2022-09-14
feat: export `metadata` object in stage 2

eduardoboucas:
Functionally, this PR adds a `metadata` object to the exports of the stage 2 file. Right now, this object contains just a `url` property containing the file path of each function, like so:

```ts
// Existing export
export const functions = {
  func1: [Function],
  func2: [Function]
}

// New export
export const metadata = {
  func1: {
    url: "file:///some/path/to/func1.ts",
  },
  func2: {
    url: "file:///some/path/to/func2.ts",
  }
}
```

This will allow the bootstrap code to match log locations to function names.

Also, it includes some housekeeping:

- Node tests were moved from `test/` to `test/node/`
- Deno tests added to `test/deno/`
- CI updated to run Deno tests
- Added stage 2 test


test/deno/stage2.test.ts
```diff
@@ -0,0 +1,57 @@
+import { assertEquals, assertStringIncludes } from 'https://deno.land/std@0.127.0/testing/asserts.ts'
+
+import { join } from 'https://deno.land/std@0.155.0/path/mod.ts'
+import { pathToFileURL } from 'https://deno.land/std@0.155.0/node/url.ts'
+
+import { getStage2Entry } from '../../deno/lib/stage2.ts'
+import { virtualRoot } from '../../deno/lib/consts.ts'
+
+Deno.test('`getStage2Entry` returns a valid stage 2 file', async () => {
```
eduardoboucas, line 9: This test was long overdue because we're dynamically generating code so it's good that we validate it as early as possible.

deno/lib/stage2.ts
```diff
@@ -14,16 +14,26 @@ const getFunctionReference = (basePath: string, func: InputFunction, index: numb
...
   const exportLines = lines.map(({ exportLine }) => exportLine).join(', ')
-  const exportDeclaration = `export const functions = {${exportLines}};`
+  const metadataLines = lines.map(({ metadata, name }) => {
+    const metadataString = JSON.stringify(metadata)
 
-  return [importLines, exportDeclaration].join('\n\n')
+    return `"${name}":${metadataString}`
+  })
+  const functionsExport = `export const functions = {${exportLines}};`
+  const metadataExport = `export const metadata = {${metadataLines}};`
```
mraerino, line 34: this scares me a bit. why are you doing this line by line instead of just calling `JSON.stringify` on a whole object?
also: did you forget a `.join(",")`?

test/deno/stage2.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  // Replacing the virtual root with the URL of the temporary directory so that
+  // we can actually import the module.
+  const normalizedStage2 = stage2.replaceAll(virtualRoot, `${baseURL.href}/`)
+
+  const stage2Path = join(directory, 'stage2.ts')
+
+  await Deno.writeTextFile(stage2Path, normalizedStage2)
+
+  const mod = await import(stage2Path)
```
mraerino, line 47: wow, importing code in the code that generated it is next level 😄

mraerino (commented):
always confused that tests are not colocated with the implementation like in Go 😅

deno/lib/stage2.ts
eduardoboucas: Great point! Updated in https://github.com/netlify/edge-bundler/pull/122/commits/71bed6300250f5e4fe97296b2602535680b9dabf.

deno/lib/stage2.ts
```diff
@@ -14,16 +14,28 @@ const getFunctionReference = (basePath: string, func: InputFunction, index: numb
...
-
-  return [importLines, exportDeclaration].join('\n\n')
+  const metadata = lines.reduce(
+    (acc, { metadata, name }) => ({
+      ...acc,
+      [name]: metadata,
+    }),
+    {},
+  )
+  const functionsExport = `export const functions = {${exportLines}};`
```
mraerino, line 35: could you do the same here if you're already on it?

khendrikse:
I saw the change of moving tests into more organised folders. I do really like structuring all of this a little bit. I remember looking into the code there are files in the src folder that are organised in folders, and some that are not. Maybe it might be interesting in the future to organise the `src` folder in general. Because it might give a little bit more context to both the code and the tests. Right now I'd almost say it might be clearer for the `deno` tests to just be a part of the `deno` folder because it'll clearly show the existing tests, and might lead to an incentive to write more when changing something there or adding something there.

Also: total nitpicking here, no prioritisation at all :)

khendrikse (approved):

test/deno/stage2.test.ts
eduardoboucas: I know, right?! 🤯

deno/lib/stage2.ts
eduardoboucas: This is not serialisable to JSON, though. The output of this will be something like:

```ts
export const functions = {"func1": func0, "func2": func1};
```

... where `func0` and `func1` are the imported modules.

eduardoboucas:
> I saw the change of moving tests into more organised folders. I do really like structuring all of this a little bit. I remember looking into the code there are files in the src folder that are organised in folders, and some that are not. Maybe it might be interesting in the future to organise the `src` folder in general. Because it might give a little bit more context to both the code and the tests. (for example, we have `bundle.ts` that is just an export of an interface, while there are other files that export functions).
> 
> Right now I'd almost say it might be clearer for the `deno` tests to just be a part of the `deno` folder because it'll clearly show the existing tests, and might lead to an incentive to write more when changing something there or adding something there.
> 
> Also: total nitpicking here, no prioritisation at all :)

Good thoughts! Let's iterate on this.

PR #126, opened 2022-09-15
feat: export `metadata` object in local stage 2

eduardoboucas:
Same as #122 but for the stage 2 file used in local development.

### DIFF
--- a/test/node/stage_2.ts
+++ b/test/node/stage_2.ts
@@ -6,6 +6,7 @@ import { pathToFileURL } from 'url'
 import test from 'ava'
 import del from 'del'
 import { execa } from 'execa'
+import semver from 'semver'
 import tmp from 'tmp-promise'
 
 import { getLocalEntryPoint } from '../../node/formats/javascript.js'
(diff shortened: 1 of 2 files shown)


test/node/stage_2.ts
```diff
@@ -0,0 +1,77 @@
...
+  // not supported. The best we can do is to naively look for some assignments.
+  if (semver.lt(process.version.slice(1), '14.0.0')) {
+    for (const func of functions) {
+      t.true(stage2.includes(`metadata["${func.name}"] = {"url":"${pathToFileURL(func.path).toString()}"}`))
+    }
+
+    return
+  }
+
+  const { stdout, stderr } = await execa('node', [stage2Path])
```
mraerino, line 64: my understanding would be that the code you are generating runs in deno? why are you executing node here?

eduardoboucas: This code runs in Node as part of local development, called by the CLI.

eduardoboucas: Called here: https://github.com/netlify/cli/blob/main/src/lib/edge-functions/proxy.js#L166-L180

eduardoboucas: Hang on, I'm holding this wrong! This is called from Node but then actually runs in Deno. 🤯 

Let me change this.

eduardoboucas: Done in https://github.com/netlify/edge-bundler/pull/126/commits/4fcd2c190c0aee5272029128b3b5b7e7363548a2. Thanks!

mraerino (approved):

PR #133, opened 2022-09-21
feat: add support for `config` export

eduardoboucas:
Adds support for in-source configuration in edge functions. Rather than forcing people to have a `netlify.toml` where they declare the paths for each edge function as well as other configuration parameters, we allow them to declare the same parameters from within a `config` export in the function's entry point file.

It works by looking for a `config` export, which should be a function. When one is found, Edge Bundler will invoke that function at build time and use its output (which must be serialisable to JSON) as the input for the configuration object.

Here's an example:

_netlify/edge-functions/hello.ts_
```ts
import { Config, Context } from "https://edge.netlify.com"

export default async (req: Request, context: Context) => {
  return new Response(`Hello, ${context.geo.country.name}`)
}

export const config: Config = () => ({
  path: "/hello"
})
```

**NOTE**: Edge functions declared using this format will run after any edge functions declared in the TOML, and they will run in alphabetical order (within the group of functions defined using this method). So if you're chaining multiple edge functions and you want to control the order in which they run, the TOML declaration method is preferred.


node/bundler.ts
```diff
@@ -124,8 +117,7 @@ const bundle = async (
   // if any.
   const importMap = new ImportMap(importMaps)
   const functions = await findFunctions(sourceDirectories)
-
-  const bundleOps = createBundleOps({
+  const functionBundle = await createBundle({
```
eduardoboucas, line 120: We're no longer creating multiple bundles, so I've simplified this a bit.

jackiewmacharia (approved):
Looks good!

PR #136, opened 2022-09-26
feat: add featureflag to allow usage of global Deno

Skn0tt:
This PR adds a new featureflag that allows usage of a build-image installed Deno binary, instead of downloading the binary during build.

cc https://github.com/netlify/pillar-workflow/issues/504

### DIFF
--- a/node/bundler.ts
+++ b/node/bundler.ts
@@ -110,6 +110,10 @@ const bundle = async (
     options.denoDir = join(cacheDirectory, 'deno_dir')
   }
 
+  if (featureFlags.edge_functions_use_global_deno) {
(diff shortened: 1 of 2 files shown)


eduardoboucas (commented):
I'm not sure I understand what this does, since `useGlobal` is [already `true` by default](https://github.com/netlify/edge-bundler/blob/main/node/bridge.ts#L59), which means Edge Bundler will prefer a global installation unless specifically told otherwise.

Skn0tt:
You are absolutely correct, and I glossed over the `?? true` bit thinking it'd surely be `?? false`. I need more coffee.

PR #138, opened 2022-09-26
feat: add support for post-cache routes

eduardoboucas:

When an edge function file doesn't have a matching declaration with a `path` property, it's deployed as a post-cache edge function.


ascorbic:
Will these run for anything that would be a 404, or for things rewritten to a functions, or somehting else?

PR #139, opened 2022-09-27
chore: replace AVA with Vitest

eduardoboucas:
Replaces AVA with Vitest.

Also moves the test files so that they're co-located with the source files (e.g. `test/node/bridge.ts` becomes `node/bridge.test.ts`).


danez (dismissed):
lgtm

jackiewmacharia (approved):

PR #140, opened 2022-09-30
feat: allow display name in manifest

danez:
Allows edge function to have a display name in the manifest. It adds `displayName` to the Declaration, which is the config that plugins write and then forwards this displayName to the edge manifest.

For https://github.com/netlify/pod-compute/issues/211


node/declaration.ts
```diff
@@ -1,10 +1,13 @@
-interface DeclarationWithPath {
+interface BaseDeclaration {
   function: string
+  displayName?: string
```
eduardoboucas, line 3: This could actually be just `name`, since we're not using that anywhere, right?

danez: tbf I think the best would be to have

```ts
{
  name: string
  displayName: string
  ...
}
```

not sure why it is called `function` in the first place. But I think even if we do not change the `function` field I think `displayName` more clearly describes what it is, simply a name that is displayed in the UI.
Having `function` and `name` is confusing to me imho

danez: Scratch what I said above. I realized that this object actually represents one route, so `function` does make sense. And the `displayName` would be the name of the route kinda? So I guess I will change `displayName` to `name`.

eduardoboucas: Yeah, exactly the `function` name makes sense because it's linking a path to a function.

Worth noting that in the future we'll be leaning more into in-source configuration, so `function` is not something that people will have to deal with. And I imagine that a `name` property there will make more sense to people than `displayName`, because it's not clear where that information will even be displayed on.

eduardoboucas (approved):

node/manifest.test.ts
```diff
@@ -30,6 +30,17 @@ test('Generates a manifest with different bundles', () => {
   expect(manifest.bundler_version).toBe(env.npm_package_version as string)
 })
 
+test('Generates a manifest with display names', () => {
+  const functions = [{ name: 'func-1', path: '/path/to/func-1.ts' }]
+  const declarations = [{ function: 'func-1', name: 'Display Name', path: '/f1/*' }]
```
eduardoboucas, line 35: Could you maybe add a second function and declaration, without a `name`, to capture the fact that we'll gracefully handle a function without one, and that it's going to simply be missing from the manifest?

eduardoboucas (approved):

node/manifest.test.ts
danez: I didn't do this because the other tests already do this: https://github.com/netlify/edge-bundler/blob/main/node/manifest.test.ts#L7-L31
Is that enough or should I still add one more case?

eduardoboucas: Good question. Personally, I tend to encapsulate all cases in the test related to the feature, as opposed to relying on the assertions from other unrelated tests. The only reason being that if you remove or change one of those tests in the future, it might not be immediately obvious to you that you're removing coverage for this other feature somewhere else in the codebase.

Totally up to you, though.

danez: okay added it to the test

eduardoboucas (approved):

khendrikse (approved):

PR #146, opened 2022-10-03
feat: update bootstrap latest to use latest version of bootstrap

khendrikse:

Updating to latest version of bootstrap

Getting https://github.com/netlify/cli/issues/4655 ready for production

### DIFF
--- a/node/formats/javascript.ts
+++ b/node/formats/javascript.ts
@@ -13,7 +13,7 @@ import { ImportMap } from '../import_map.js'
 import type { FormatFunction } from '../server/server.js'
 import { getFileHash } from '../utils/sha256.js'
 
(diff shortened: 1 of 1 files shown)


biruwon (dismissed):

node/formats/javascript.ts
```diff
@@ -118,7 +118,7 @@ const getLocalEntryPoint = (
     return `
       try {
         const { default: func } = await import("${url}");
-    
+
```
eduardoboucas, line 121: Wow, how did this happen? Is the linter no running in the CI? 🤔

eduardoboucas (approved):
We should rename this to `feat:` or `fix:` before merging, otherwise a release won't be generated.

PR #161, opened 2022-10-13
feat: support JSX and TSX in edge functions

ascorbic:

Adds support for .jsx and .tsx entrypoints in edge functions

Deno includes out of the box support for jsx and tsx. However edge bundler doesn't recognise these files, so they can't be used for edge functions at the moment. This PR adds .tsx and .jsx to the list of allowed extensions. It allows things like this:

```tsx
import React from "https://esm.sh/react@18";
import { renderToReadableStream } from "https://esm.sh/v96/react-dom@18.2.0/deno/server.js";

export default async function handler(req: Request) {
  const url = new URL(req.url);
  const username = url.searchParams.get("username");

  if (!username) {
    url.searchParams.set("username", "netlify");
    return Response.redirect(url.toString(), 302);
  }

  const controller = new AbortController();
  let didError = false;
  try {
    const stream = await renderToReadableStream(
      <html>
        <title>Hello</title>
        <body>
          <h1>Hello {username}</h1>
        </body>
      </html>,
      {
        signal: controller.signal,
        onError(error: unknown) {
          didError = true;
          console.error(error);
        },
      },
    );

    return new Response(stream, {
      status: didError ? 500 : 200,
      headers: { "Content-Type": "text/html" },
    });
  } catch (error) {
    return new Response(
      "<!doctype html><p>Something went wrong</p>",
      {
        status: 500,
        headers: { "Content-Type": "text/html" },
      },
    );
  }
}
```

Importing a JSX file into a separate js or ts entrypoint

**Checklist**

so im basically very smol

### DIFF
--- a/node/finder.ts
+++ b/node/finder.ts
@@ -8,9 +8,16 @@ const ALLOWED_EXTENSIONS = new Set(['.js', '.jsx', '.ts', '.tsx'])
 
 const findFunctionInDirectory = async (directory: string): Promise<EdgeFunction | undefined> => {
   const name = basename(directory)
-  const candidatePaths = [`${name}.js`, `index.js`, `${name}.ts`, `index.ts`].map((filename) =>
-    join(directory, filename),
-  )
+  const candidatePaths = [
+    `${name}.js`,
+    `index.js`,
+    `${name}.ts`,
+    `index.ts`,
+    `${name}.jsx`,
+    `index.jsx`,
+    `${name}.tsx`,
+    `index.tsx`,
+  ].map((filename) => join(directory, filename))
 
   let functionPath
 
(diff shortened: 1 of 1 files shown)


ascorbic:
Aaah, it works locally because it's not in a directory. I'll update this

node/finder.ts
```diff
@@ -4,13 +4,20 @@ import { basename, extname, join } from 'path'
...
+  const candidatePaths = [
+    `${name}.js`,
+    `index.js`,
+    `${name}.ts`,
+    `index.ts`,
+    `${name}.jsx`,
+    `index.jsx`,
+    `${name}.tsx`,
+    `index.tsx`,
+  ].map((filename) => join(directory, filename))
```
eduardoboucas, line 20: How do you feel about using `ALLOWED_EXTENSIONS` to generate all candidate paths, as opposed to enumerating them?

```suggestion
  const candidatePaths = [...ALLOWED_EXTENSIONS]
    .flatMap((extension) => [name + extension, `index${extension}`])
    .map((filename) => join(directory, filename))
```

ascorbic: Yeah, that is better

eduardoboucas (approved):
✨

PR #165, opened 2022-10-19
feat: add support for post-cache routes

eduardoboucas:

Looks for a `mode` property in a function declaration (either in the TOML or in-source) and if its value is `after_cache`, the route is added to `post_cache_routes` instead of `routes` in the manifest file. This makes the function run after the cache.

Internal: https://github.com/netlify/pod-compute/issues/208

### DIFF
--- a/node/config.test.ts
+++ b/node/config.test.ts
@@ -147,7 +147,7 @@ test('Ignores function paths from the in-source `config` function if the feature
   })
   const generatedFiles = await fs.readdir(tmpDir.path)
 
-  expect(result.functions.length).toBe(4)
+  expect(result.functions.length).toBe(6)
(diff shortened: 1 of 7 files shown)


khendrikse (approved):

eduardoboucas:
After looking at existing configuration properties in https://docs.netlify.com/configure-builds/file-based-configuration/, I feel that using kebab case in the `mode` values is more appropriate than snake case, so I've changed that in https://github.com/netlify/edge-bundler/pull/165/commits/fdc4043959abead7f99f8b88ea8327fe086cd4c5.

Can I get another ✅ , please?

khendrikse (approved):

ascorbic:
😍

PR #174, opened 2022-10-26
fix: run `config` export with `--allow-env`

eduardoboucas:

When invoking the edge function at build time to resolve the `config` export, we should call the Deno CLI with `--allow-env`, otherwise we'll get an error if the function is trying to access `Deno.env` in the global scope.


jackiewmacharia (approved):

ascorbic (approved):
🚢

PR #195, opened 2022-11-15
feat: add detection + nice error message for npm import errors

Skn0tt:
Closes https://github.com/netlify/pod-compute/issues/287

### DIFF
--- a/node/bundler.test.ts
+++ b/node/bundler.test.ts
@@ -176,7 +176,7 @@ test('Prints a nice error message when user tries importing NPM module', async (
     expect(error).toBeInstanceOf(NPMImportError)
     expect((error as NPMImportError).moduleName).toEqual('p-retry')
     expect((error as NPMImportError).message).toEqual(
-      `It seems like you're trying to import an NPM module. This is only supported in Deno via CDNs like esm.sh, have you tried 'import ... from "https://esm.sh/p-retry"'?`,
+      `It seems like you're trying to import an npm module. This is only supported in Deno via CDNs like esm.sh. Have you tried 'import mod from "https://esm.sh/p-retry"'?`,
     )
   }
 })
--- a/node/bundler.test.ts
+++ b/node/bundler.test.ts
@@ -154,7 +154,7 @@ test('Adds a custom error property to user errors during bundling', async () => 
   }
 })
 
-test.only('Prints a nice error message when user tries importing NPM module', async () => {
+test('Prints a nice error message when user tries importing NPM module', async () => {
   expect.assertions(3)
 
   const sourceDirectory = resolve(fixturesDir, 'imports_npm_module', 'functions')
--- a/node/bundler.test.ts
+++ b/node/bundler.test.ts
@@ -11,6 +11,7 @@ import { fixturesDir } from '../test/util.js'
 
 import { BundleError } from './bundle_error.js'
 import { bundle, BundleOptions } from './bundler.js'
+import { NPMImportError } from './npm_import_error.js'
 
 test('Produces a JavaScript bundle and a manifest file', async () => {
   const sourceDirectory = resolve(fixturesDir, 'with_import_maps', 'functions')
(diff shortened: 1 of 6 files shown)


node/bundler.test.ts
```diff
@@ -153,6 +154,29 @@ test('Adds a custom error property to user errors during bundling', async () =>
...
+    },
+  ]
+
+  try {
+    await bundle([sourceDirectory], tmpDir.path, declarations)
+  } catch (error) {
+    expect(error).toBeInstanceOf(NPMImportError)
+    expect((error as NPMImportError).moduleName).toEqual('p-retry')
+    expect((error as NPMImportError).message).toEqual(
+      `It seems like you're trying to import an NPM module, did you mean to 'import ... from "npm:p-retry"'?`,
```
eduardoboucas, line 175: We shouldn't offer this suggestion because don't support it.

node/npm_import_error.ts
```diff
@@ -0,0 +1,23 @@
...
+    this.stack = originalError.stack
+
+    // https://github.com/microsoft/TypeScript-wiki/blob/8a66ecaf77118de456f7cd9c56848a40fe29b9b4/Breaking-Changes.md#implicit-any-error-raised-for-un-annotated-callback-arguments-with-no-matching-overload-arguments
+    Object.setPrototypeOf(this, NPMImportError.prototype)
+  }
+}
+
+const checkNpmImportError = (input: unknown): NPMImportError | undefined => {
+  if (input instanceof Error) {
+    const match = input.message.match(/Relative import path "(.*)" not prefixed with/)
```
eduardoboucas, line 15: Matching the error message feels a bit brittle. As I suggested [here](https://github.com/netlify/pod-compute/issues/287#issuecomment-1314520156), we could leverage the fact that we're in control of the module resolving (in production anyway, where we use ESZIP).

We should be able to look at a specifier and see whether it's a full URL, a relative path, or anything else. If it's _anything else_ and the loading fails, I think we can safely infer that people are either trying to load an npm module or a specifier that should be in the import map but isn't.

I think doing that check in the loader would be more robust.

eduardoboucas: Ah, interesting! I thought the ESZIP loader would be called even for paths that don't match a URL format or a path. In that case, ignore me!

node/npm_import_error.ts
```diff
@@ -0,0 +1,25 @@
+class NPMImportError extends Error {
+  constructor(originalError: Error, public readonly moduleName: string) {
+    super(
+      `It seems like you're trying to import an NPM module. This is only supported in Deno via CDNs like esm.sh, have you tried 'import ... from "https://esm.sh/${moduleName}"'?`,
```
eduardoboucas, line 4: A couple of nits:

- I think "npm" is the correct capitalisation
- Changed `import ...` to `import mod` because that's a valid instruction, in case people try to copy/paste

```suggestion
      `It seems like you're trying to import an npm module. This is only supported in Deno via CDNs like esm.sh. Have you tried 'import mod from "https://esm.sh/${moduleName}"'?`,
```

node/formats/eszip.ts
```diff
@@ -52,6 +53,11 @@ const bundleESZIP = async ({
   try {
     await deno.run(['run', ...flags, bundler, JSON.stringify(payload)], { pipeOutput: true })
   } catch (error: unknown) {
+    const npmImportError = checkNpmImportError(error)
+    if (npmImportError) {
+      throw npmImportError
```
eduardoboucas, line 58: `wrapBundleError` is responsible for setting the `customErrorInfo` property, which is then used to distinguish user errors from system errors. By not wrapping `npmImportError`, we'll start incorrectly treating these as system errors.

Skn0tt: that's good to know! fixed in https://github.com/netlify/edge-bundler/pull/195/commits/d49211a50796209f346379a583ae972070c32230 and added a comment about that in https://github.com/netlify/edge-bundler/pull/195/commits/fd38b1f6f7e9a74cc983694312184ccc64a8bda8

node/formats/eszip.ts
```diff
@@ -51,7 +52,10 @@ const bundleESZIP = async ({
 
   try {
     await deno.run(['run', ...flags, bundler, JSON.stringify(payload)], { pipeOutput: true })
-  } catch (error: unknown) {
+  } catch (_error: unknown) {
+    let error = _error
+    error = checkNpmImportError(error) ?? error
```
eduardoboucas, line 57: I find this sequence confusing, to be honest.

I think a method called `checkNpmImportError` that returns either an error or `undefined` isn't the most natural?

I would either:

- Call it `isNpmImportError` and make it return a boolean — you'd then need to call the `NPMImportError` separately if that condition is true, or
- Call it something like `wrapNpmImportError` which always returns an error — either the original error if it's not an npm import error, or an instance of `NPMImportError` if it is

What do you think?

node/npm_import_error.ts
```diff
@@ -0,0 +1,25 @@
+class NPMImportError extends Error {
+  constructor(originalError: Error, moduleName: string) {
+    super(
+      `It seems like you're trying to import an npm module. This is only supported in Deno via CDNs like esm.sh. Have you tried 'import mod from "https://esm.sh/${moduleName}"'?`,
```
eduardoboucas, line 4: Btw, extracting the module name with the regular expression and showing it here is a really nice touch! ✨

node/formats/eszip.ts
Skn0tt: Agreed! looks like I write too much go :D changed in https://github.com/netlify/edge-bundler/pull/195/commits/67475ddfd2de1e72bc69bca48e7545eed3defd0e

Skn0tt:
we should really install a pre-commit hook with prettier ...

eduardoboucas (approved):

PR #200, opened 2022-11-16
feat: validate edge function signature at build time

jackiewmacharia:

This fixes https://github.com/netlify/pod-compute/issues/249

**Checklist**

### DIFF
--- a/deno/config.ts
+++ b/deno/config.ts
@@ -11,6 +11,14 @@ try {
   Deno.exit(exitCodes.ImportError)
 }
 
+if (func.default === undefined) {
+  Deno.exit(exitCodes.NoDefaultExport)
+}
+
+if (typeof func.default !== 'function') {
+  Deno.exit(exitCodes.InvalidDefaultExport)
+}
+
 if (func.config === undefined) {
   Deno.exit(exitCodes.NoConfig)
 }
--- a/deno/config.ts
+++ b/deno/config.ts
@@ -11,10 +11,6 @@ try {
   Deno.exit(exitCodes.ImportError)
 }
 
-if (func.default === undefined) {
-  Deno.exit(exitCodes.NoDefaultExport)
-}
-
 if (typeof func.default !== 'function') {
   Deno.exit(exitCodes.InvalidDefaultExport)
 }
--- a/node/config.test.ts
+++ b/node/config.test.ts
@@ -281,17 +281,17 @@ test('Fails validation if default export is not function', async () => {
 
   await fs.writeFile(path, func.source)
 
-  expect(async () => {
-    await getFunctionConfig(
-      {
-        name: func.name,
-        path,
-      },
-      new ImportMap([importMapFile]),
-      deno,
-      logger,
-    )
-  }).toThrowError(defaultExportNotFunctionErr)
+  const config = await getFunctionConfig(
+    {
+      name: func.name,
+      path,
+    },
+    new ImportMap([importMapFile]),
+    deno,
+    logger,
+  )
+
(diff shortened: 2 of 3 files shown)


deno/config.ts
```diff
@@ -11,6 +11,14 @@ try {
   Deno.exit(exitCodes.ImportError)
 }
 
+if (func.default === undefined) {
```
eduardoboucas, line 14: I understand that this level of granularity gives people more information — i.e. no export vs. export with wrong type. But do you think it's worth the additional logic for the extra branch? If we just do the `typeof` check and show the same message for the two cases with a `ntl.fyi` link pointing to https://docs.netlify.com/edge-functions/api/, I think that would achieve the same goal with half the logic.

What do you think?

jackiewmacharia: I think it makes sense, I'll implement it.

node/config.ts
```diff
@@ -121,6 +124,12 @@ const logConfigError = (func: EdgeFunction, exitCode: number, stderr: string, lo
 
       break
 
+    case ConfigExitCode.InvalidDefaultExport:
+      errMsg = `Default export in edge function at '${func.path}' must be a function. More on the Edge Functions API at https://docs.netlify.com/edge-functions/api/`
```
jackiewmacharia, line 128: This needs a `ntl.fyi` shortened link

deno/config.ts
jackiewmacharia: This is now done _(waiting on a `ntl.fyi` link though)_.

node/config.ts
```diff
@@ -18,6 +18,8 @@ enum ConfigExitCode {
   InvalidExport,
   RuntimeError,
   SerializationError,
+  NoDefaultExport,
```
eduardoboucas, line 21: I think this is not needed anymore?

node/config.ts
```diff
@@ -121,6 +124,12 @@ const logConfigError = (func: EdgeFunction, exitCode: number, stderr: string, lo
 
       break
 
+    case ConfigExitCode.InvalidDefaultExport:
+      errMsg = `Default export in edge function at '${func.path}' must be a function. More on the Edge Functions API at https://ntl.fyi/edge-api`
```
eduardoboucas, line 128: Do we need to initialise this variable at the top and mutate it here? Could we instead just send the string directly to the `Error` constructor?

node/config.ts
```diff
@@ -121,6 +124,12 @@ const logConfigError = (func: EdgeFunction, exitCode: number, stderr: string, lo
...
+      throw new Error(errMsg)
+
+      break
```
eduardoboucas, line 131: This is unreachable code since we're throwing above.

node/config.test.ts
```diff
@@ -21,6 +21,9 @@ const importMapFile = {
   },
 }
 
+const invalidDefaultExportErr = (path: string) =>
+  `Default export in edge function at '${path}' must be a function. More on the Edge Functions API at https://ntl.fyi/edge-api`
```
eduardoboucas, line 25: Small nit that you're free to ignore, but I feel that we could make the message shorter by not repeating "edge function".

```suggestion
  `Default export in '${path}' must be a function. More on the Edge Functions API at https://ntl.fyi/edge-api.`
```

deno/config.ts
jackiewmacharia: now done done

eduardoboucas (commented):
I left a few small comments, but otherwise looking great!

eduardoboucas (approved):
🚀

PR #201, opened 2022-11-17
feat: add `layer:` prefix to ESZIP loader

eduardoboucas:
This should've been part of #198, but I forgot. We need to tell the ESZIP loader to treat any specifiers that start with `layer:` as dangling imports.


Skn0tt (dismissed):

eduardoboucas:
@Skn0tt I realised I was missing the `layers` property in the `bundle` entry point. Also added a new test. Could I get another ✅, please? 🙏🏻

khendrikse (approved):

PR #209, opened 2022-11-18
fix: tag edge function signature errors as user errors

jackiewmacharia:

Example: This fixes #5012

**Checklist**

### DIFF
--- a/node/config.ts
+++ b/node/config.ts
@@ -5,6 +5,7 @@ import { pathToFileURL } from 'url'
 import tmp from 'tmp-promise'
 
 import { DenoBridge } from './bridge.js'
+import { wrapBundleError } from './bundle_error.js'
 import { EdgeFunction } from './edge_function.js'
 import { ImportMap } from './import_map.js'
 import { Logger } from './logger.js'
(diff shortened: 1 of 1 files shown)


node/config.ts
```diff
@@ -123,8 +124,12 @@ const logConfigError = (func: EdgeFunction, exitCode: number, stderr: string, lo
...
 
     case ConfigExitCode.InvalidDefaultExport:
-      throw new Error(
-        `Default export in '${func.path}' must be a function. More on the Edge Functions API at https://ntl.fyi/edge-api.`,
+      throw wrapBundleError(
+        new Error(
+          `Default export in '${func.path}' must be a function. More on the Edge Functions API at https://ntl.fyi/edge-api.`,
+        ),
+        // empty string because it's not associated with a bundling format
+        { format: '' },
```
eduardoboucas, line 132: Rather than specifying this, could we make the options parameter in `wrapBundleError` optional?

I would go as far as making the options parameter in the `BundleError` constructor itself optional, and I'd call it directly from here:

```ts
throw new BundleError(`Default export in '${func.path}' must be a function. More on the Edge Functions API at https://ntl.fyi/edge-api.`)
```

I think the purpose of `wrapBundleError` is to wrap something that may or may not be an error. In this case we know for a fact that we're dealing with an error because we're the ones creating it.

jackiewmacharia: done

eduardoboucas (approved):
✨

PR #220, opened 2022-11-23
fix: deep merge in source and toml config

jackiewmacharia:

This relates to https://github.com/netlify/pod-compute/issues/292

**Checklist**

### DIFF
--- a/node/declaration.ts
+++ b/node/declaration.ts
@@ -25,7 +25,8 @@ export const getDeclarationsFromConfig = (
 
   // We start by iterating over all the TOML declarations. For any declaration
   // for which we also have a function configuration object, we replace the
-  // path or cache because that object takes precedence.
+  // defined config (currently path or cache or both) because that object takes
+  // precedence.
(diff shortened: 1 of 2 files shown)


node/declaration.ts
```diff
@@ -25,14 +25,15 @@ export const getDeclarationsFromConfig = (
...
   // for which we also have a function configuration object, we replace the
-  // path because that object takes precedence.
+  // defined config (currently path or cache or both) because that object takes
+  // precedence.
   for (const declaration of tomlDeclarations) {
-    const { path } = functionsConfig[declaration.function] ?? {}
+    const config = functionsConfig[declaration.function] ?? {}
 
-    if (path) {
+    if (config) {
```
eduardoboucas, line 33: I don't think the `else` branch will ever be reached, because `config` has a fallback value of `??`. This wasn't the case before, since we were checking the `path` property specifically.

Or am I holding it wrong?

eduardoboucas (approved):

PR #223, opened 2022-11-23
feat: add custom layer specifiers to import maps

eduardoboucas:

This PR creates an import map for any custom layer specifiers before we try to extract the in-source configuration, otherwise `deno run` will throw.


eduardoboucas:
Actually, I'll address this differently.

PR #224, opened 2022-11-23
feat!: remove support for JavaScript bundles

eduardoboucas:

We've been generating ESZIP bundles for all sites for a while now. This PR formally removes support for JavaScript bundles.


jackiewmacharia (approved):
🚀

PR #226, opened 2022-11-24
fix: generate manifest with declaration pattern instead of compiled regex

orinokai:

The Next Runtime uses a manifest file generated by Next to create edge function definitions. Part of the those definitions is a route that is defined by a regex pattern in the manifest file.

Next 13 has started using named capture groups in those patterns, which have a slightly different syntax in Golang. In the Next Runtime we convert these patterns to Golang compatible syntax, however the Edge Bundler currently attempts to compile these regex patterns before bundling, causing an error.

https://github.com/netlify/next-runtime/pull/1809

I couldn't spot any reason why these patterns needed to be compiled into regex objects, so this PR simply keeps the patterns as strings, avoiding the error.

An alternative solution would be to perform the Golang syntax transformation in the edge-bundler, later in the bundling stage, instead of doing it in the Next runtime. This could potentially be a more generic solution, but I don't know of the wider impact of this option.

### DIFF
--- a/node/manifest.ts
+++ b/node/manifest.ts
@@ -45,8 +45,8 @@ const generateManifest = ({ bundles = [], declarations = [], functions, layers =
       return
     }
 
-    const pattern = getRegularExpression(declaration)
-    const serializablePattern = pattern.source.replace(/\\\//g, '/')
+    const pattern = getDeclarationPattern(declaration)
+    const serializablePattern = pattern.replace(/\\\//g, '/')
     const route = {
       function: func.name,
       name: declaration.name,
@@ -74,9 +74,9 @@ const generateManifest = ({ bundles = [], declarations = [], functions, layers =
   return manifest
 }
 
-const getRegularExpression = (declaration: Declaration) => {
+const getDeclarationPattern = (declaration: Declaration) => {
   if ('pattern' in declaration) {
-    return new RegExp(declaration.pattern)
+    return declaration.pattern
   }
 
   // We use the global flag so that `globToRegExp` will not wrap the expression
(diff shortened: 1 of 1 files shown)


ascorbic:
tbh, I still think this is worth having, even if you've worked around it now. It should allow golang regexes because that's what it uses.

orinokai:
reopening as per @ascorbic's comment above

eduardoboucas (approved):
LGTM

eduardoboucas:
Actually, there's something that we get from creating a `RegExp` object, which is validating the regular expressions. We'd lose that if we start accepting any string.

If we stop creating a `RegExp`, I think we need to add some additional validation to ensure we're always populating the manifest with valid expressions.

eduardoboucas:
Yeah, I agree that validating it against a Go regular expression would be best, and I'm happy to drop the `RegExp` creation at that point. What I don't want is for us to remove one before adding the other.

orinokai:
I think we would need to parse the pattern to validate it successfully - as far as I know it's not possible to validate a regex with a regex. Golang uses RE2 syntax, I believe, so we could use a library such as https://www.npmjs.com/package/re2 to do so?

eduardoboucas:
I think this is a bit problematic: https://github.com/uhop/node-re2#precompiled-artifacts

Any dependencies we add to Edge Bundler will become transitive dependencies of both Netlify Build and the CLI, and adding one with native bindings is not something we should do lightly.

We don't need the full power of that library, since we're just interested in validating a regex, not actually running it. I _believe_ y'all have been parsing them into ASTs — is that something that we could leverage here?

ascorbic:
We do parse them, but I think that's going to have the same problem: it will only parse JavaScript regexes

orinokai:
Closing in favour of #298

PR #230, opened 2022-11-30
feat: support for multiple paths in in source configuration

jackiewmacharia:

https://github.com/netlify/pod-compute/issues/295

This PR adds support for multiple paths in ISC. It allows for path to be either a string or array of strings.

**Checklist**

cute animal


node/declaration.ts
```diff
@@ -32,17 +34,48 @@ export const getDeclarationsFromConfig = (
...
+    // If no config is found, add the declaration as is
+    if (Object.keys(config).length === 0) {
+      declarations.push(declaration)
+    }
+
+    // log if the path property is not an array
+    if (config.path && !Array.isArray(config.path)) {
+      log.user(
+        `Can't use in source config path property: ${config.path}. Function configuration path property must be an array.`,
+      )
```
danez, line 46: Not sure how far we have rolled out ISC for edge functions yet, but do we need to provide backwards compatibility? So if path is a string, simply convert it to an array and print a warning?

jackiewmacharia: This is now done.

node/declaration.test.ts
```diff
@@ -16,12 +16,13 @@ test('In source config takes precedence over netlify.toml config', () => {
   ]
 
   const funcConfig = {
-    geolocation: { path: '/geo-isc', cache: 'manual' },
-    json: { path: '/json', cache: 'off' },
+    geolocation: { path: ['/geo-isc', '/*'], cache: 'manual' },
+    json: { path: ['/json'], cache: 'off' },
```
eduardoboucas, line 20: Can we maybe leave the `path` for `json` as a string to cover that flow?

node/declaration.test.ts
```diff
@@ -75,3 +77,53 @@ test('In source config works independent of the netlify.toml file if a path is d
...
+
+  expect(getDeclarationsFromConfig(tomlConfig, funcConfig, deployConfig)).toEqual(expectedDeclarations)
+})
+
+test("In source config path property works if it's not an array", () => {
+  const tomlConfig = [{ function: 'json', path: '/json-toml', cache: 'off' }]
+
+  const funcConfig = {
+    json: { path: '/json', cache: 'manual' },
+  } as Record<string, FunctionConfig>
```
eduardoboucas, line 98: Do you need the `as` here?

node/declaration.test.ts
khendrikse: ✅

test/fixtures/with_config/.netlify/edge-functions/framework-func1.ts
```diff
@@ -7,5 +7,5 @@ export default async () => {
 }
 
 export const config = () => ({
-  path: '/framework-func1',
```
eduardoboucas, line 10: Is there a reason to change all of these to the array format? Also, there's something weird going on with the formatting (see the missing space after `path:`).

node/declaration.ts
```diff
@@ -30,19 +30,44 @@ export const getDeclarationsFromConfig = (
...
+      // If we have path specified create a declaration for each path
+    } else if (config.path?.length) {
+      // eslint-disable-next-line max-depth
+      if (!Array.isArray(config.path)) config.path = [config.path]
+
+      config.path.forEach((path) => {
+        declarations.push({ ...declaration, ...config, path })
+      })
+
+      // If only cache was specified, add it to the declaration
```
eduardoboucas, line 48: I don't think this comment is accurate anymore, since we're not targeting the `cache` property specifically?

node/declaration.ts
```diff
@@ -30,19 +30,44 @@ export const getDeclarationsFromConfig = (
...
   for (const declaration of [...tomlDeclarations, ...deployConfig.declarations]) {
-    const config = functionsConfig[declaration.function] ?? {}
+    const config = functionsConfig[declaration.function]
+
+    // If no config is found, add the declaration as is
+    if (!config) {
+      declarations.push(declaration)
+
+      // If we have path specified create a declaration for each path
+    } else if (config.path?.length) {
```
eduardoboucas, line 40: I think `config.path` may be a string at this point? When that happens, do we want to be checking its `length`?

khendrikse:
@eduardoboucas sorry all my comments weren't sent because I did not review correctly :p

node/declaration.ts
khendrikse: ✅

node/declaration.test.ts
khendrikse: Sorry, my comment came in late, can you explain you wondering its necessity?

test/fixtures/with_config/.netlify/edge-functions/framework-func1.ts
khendrikse: I'll change them back to a string, as I think a string will probably still be used more often, and we have tests in place to make sure an array would work as well.

node/declaration.test.ts
```diff
@@ -75,3 +77,53 @@ test('In source config works independent of the netlify.toml file if a path is d
   const declarationsWithoutISCPath = getDeclarationsFromConfig(tomlConfig, funcConfigWithoutPath, deployConfig)
   expect(declarationsWithoutISCPath).toEqual(expectedDeclarationsWithoutISCPath)
 })
+
+test('In source config works if only the cache config property is set', () => {
```
eduardoboucas, line 81: Mega nit.

```suggestion
test('In-source config works if only the cache config property is set', () => {
```

node/declaration.test.ts
```diff
@@ -75,3 +77,53 @@ test('In source config works independent of the netlify.toml file if a path is d
...
+  const funcConfig = {
+    geolocation: { cache: 'manual' },
+  } as Record<string, FunctionConfig>
+
+  const expectedDeclarations = [{ function: 'geolocation', path: '/geo', cache: 'manual' }]
+
+  expect(getDeclarationsFromConfig(tomlConfig, funcConfig, deployConfig)).toEqual(expectedDeclarations)
+})
+
+test("In source config path property works if it's not an array", () => {
```
eduardoboucas, line 93: Mega nit.

```suggestion
test("In-source config path property works if it's not an array", () => {
```

node/declaration.test.ts
```diff
@@ -75,3 +77,53 @@ test('In source config works independent of the netlify.toml file if a path is d
...
+  const funcConfig = {
+    json: { path: '/json', cache: 'manual' },
+  } as Record<string, FunctionConfig>
+
+  const expectedDeclarations = [{ function: 'json', path: '/json', cache: 'manual' }]
+
+  expect(getDeclarationsFromConfig(tomlConfig, funcConfig, deployConfig)).toEqual(expectedDeclarations)
+})
+
+test("In source config path property works if it's not an array and it's not present in toml or deploy config", () => {
```
eduardoboucas, line 105: Mega nit.

```suggestion
test("In-source config path property works if it's not an array and it's not present in toml or deploy config", () => {
```

node/declaration.test.ts
```diff
@@ -75,3 +77,53 @@ test('In source config works independent of the netlify.toml file if a path is d
...
+
+  const expectedDeclarations = [
+    { function: 'geolocation', path: '/geo', cache: 'off' },
+    { function: 'json', path: '/json-isc', cache: 'manual' },
+  ]
+
+  expect(getDeclarationsFromConfig(tomlConfig, funcConfig, deployConfig)).toEqual(expectedDeclarations)
+})
+
+test('In source config works if path property is an empty array with cache value specified', () => {
```
eduardoboucas, line 119: Mega nit.

```suggestion
test('In-source config works if path property is an empty array with cache value specified', () => {
```

node/declaration.ts
```diff
@@ -30,19 +30,44 @@ export const getDeclarationsFromConfig = (
...
+      // eslint-disable-next-line max-depth
+      if (!Array.isArray(config.path)) config.path = [config.path]
```
eduardoboucas, line 42: Rather than mutating config here, I think it would be nicer to create a new variable with the normalised values. Something like:

```js
const paths = Array.isArray(config.path) ? config.path : [config.path]
```

I _think_  that should also let you get rid of the `eslint-disable-next-line` directive.

node/declaration.ts
eduardoboucas: Hmm I understand your rationale, but I don't think it's immediately obvious. Perhaps adding that context to the comment would help?

node/declaration.ts
```diff
@@ -30,19 +30,44 @@ export const getDeclarationsFromConfig = (
...
+
+      config.path.forEach((path) => {
+        declarations.push({ ...declaration, ...config, path })
+      })
+
+      // If a config was specified, add it to the declaration
+    } else {
```
eduardoboucas, line 49: I'm not sure I understand this branch. Does this happen when there's an in-source config block with an empty path?

By the way, if you want to avoid disabling the ESLint rule, you could do:

```js
declarations.push({ path, ...declaration, ...rest })
```

node/declaration.ts
```diff
@@ -30,19 +30,44 @@ export const getDeclarationsFromConfig = (
...
   }
 
   // Finally, we must create declarations for functions that are not declared
   // in the TOML at all.
   for (const name in functionsConfig) {
-    const { path, ...config } = functionsConfig[name]
+    const { ...config } = functionsConfig[name]
+    let { path } = functionsConfig[name]
 
+    // If we have path specified create a declaration for each path
```
eduardoboucas, line 64:
```suggestion
    // If we have a path specified, create a declaration for each path
```

node/declaration.ts
```diff
@@ -30,19 +30,44 @@ export const getDeclarationsFromConfig = (
...
     if (!functionsVisited.has(name) && path) {
-      declarations.push({ ...config, function: name, path })
+      if (!Array.isArray(path)) path = [path]
```
eduardoboucas, line 66: Same suggestion as above:

```suggestion
      const paths = Array.isArray(path) ? path : [path]
```

node/declaration.ts
khendrikse: Shall do!

node/declaration.ts
khendrikse: Nice, I wanted to work on getting rid of the mutations yesterday but my brain was a bit foggy, I like this change a lot

node/declaration.test.ts
khendrikse: @eduardoboucas do you still want me to address this one?

eduardoboucas: Sorry, I had missed your question. I thought the type could be inferred, but I can see that's not the case. Ignore me!

node/declaration.ts
```diff
@@ -30,19 +30,45 @@ export const getDeclarationsFromConfig = (
...
   // takes precedence.
   for (const declaration of [...tomlDeclarations, ...deployConfig.declarations]) {
-    const config = functionsConfig[declaration.function] ?? {}
+    const config = functionsConfig[declaration.function]
+
+    // If no config is found, add the declaration as is
+    if (!config) {
+      declarations.push(declaration)
+
+      // If we have a path specified as either a string or non-tempty array
```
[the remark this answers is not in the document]
khendrikse: But the array is not tempting Eduardo. :p

node/declaration.ts
eduardoboucas: Are you still planning on doing this?

khendrikse: :D @eduardoboucas guess who forgot to push her changes, it's me! 

Should be up to date now

eduardoboucas (approved):
✨ Awesome work, Jackie and Karin! ❤️

PR #235, opened 2022-12-05
feat: save import map URL to manifest

eduardoboucas:

Import maps aren't working as expected at the moment for a couple of reasons.

First of all, we're not setting the `import_map_url` property in the isolate config, because we're not even writing that to the manifest.

Also, we're writing absolute paths (from the host machine in the build system) into the import map, instead of converting those to ESZIP paths using the virtual root prefix.

This PR addresses that.

Closes https://github.com/netlify/pod-compute/issues/313.

### DIFF
--- a/node/import_map.ts
+++ b/node/import_map.ts
@@ -1,6 +1,6 @@
 import { Buffer } from 'buffer'
 import { promises as fs } from 'fs'
-import { dirname, isAbsolute, relative } from 'path'
+import { dirname, isAbsolute, posix, relative, sep } from 'path'
 import { fileURLToPath, pathToFileURL } from 'url'
 
 import { parse } from '@import-maps/resolve'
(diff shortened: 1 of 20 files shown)


test/util.ts
```diff
@@ -10,4 +14,20 @@ const url = new URL(import.meta.url)
 const dirname = fileURLToPath(url)
 const fixturesDir = resolve(dirname, '..', 'fixtures')
 
-export { fixturesDir, testLogger }
+const useFixture = async (fixtureName: string) => {
```
eduardoboucas, line 17: I created this helper and refactored all tests in `bundler.test.ts` to use it because I realised our tests were not mimicking the structure observed in production. In production, the base path (e.g. `/project`) is a parent to both the source directory (`/project/netlify/edge-functions`) and the dist directory (`/project/.netlify/edge-functions-dist`), but in tests we were using as dist directory a temporary path that is something like `/var/tmp`.

With this helper, we copy the source directory into that temporary directory and make the base path.

node/import_map.ts
```diff
@@ -28,22 +28,50 @@ class ImportMap {
...
+    return { ...parsedImportMap, imports: newImports }
   }
 
   add(file: ImportMapFile) {
     this.files.push(file)
   }
 
-  getContents() {
-    let imports: Record<string, URL | null> = {}
+  getContents(rootPath?: string) {
```
danez, line 70: Shouldn't rootPath be mandatory? For example `. toDataURL()` would now create a different import map?

eduardoboucas: We can't have relative paths when the import map is described as a data URL, so in that situation we'll leave the paths as absolute.

PR #241, opened 2022-12-06
fix: correct import_map attribute name in validation

Skn0tt:
Follow up to https://github.com/netlify/edge-bundler/pull/239, where the schema said `importMapURL` as the key name instead of `import_map` which is correct.

### DIFF
--- a/node/bundler.test.ts
+++ b/node/bundler.test.ts
@@ -11,6 +11,7 @@ import { useFixture } from '../test/util.js'
 import { BundleError } from './bundle_error.js'
 import { bundle, BundleOptions } from './bundler.js'
 import { isNodeError } from './utils/error.js'
+import { validateManifest } from './validation/manifest/index.ts'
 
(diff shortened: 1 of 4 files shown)


Skn0tt:
on my machine, I have do set `FORCE_COLOR=0` for the snapshot testing to work correctly. Is it the same for you, @eduardoboucas? If I don't, it outputs a bunch of colour escape codes into the snapshot.

eduardoboucas (approved):

eduardoboucas:
> on my machine, I have do set `FORCE_COLOR=0` for the snapshot testing to work correctly. Is it the same for you, @eduardoboucas? If I don't, it outputs a bunch of colour escape codes into the snapshot.

You can also use `npm run test:dev:vitest:watch`.

Skn0tt:
good to know! now I see that `package.json` also has `FORCE_COLOR=0`, that makes sense :)

PR #256, opened 2022-12-13
feat!: add support for user-defined import maps

eduardoboucas:

Currently, the only way to use import maps with edge functions is by placing an import map file inside the internal edge functions directory, which is meant for integrations.

This PR adds support for user-defined import maps by looking for a [Deno configuration file](https://deno.land/manual@v1.28.3/getting_started/configuration_file) at `/deno.json` or `/deno.jsonc`. When this file is found and it contains an `importMap` property, we load that import map into the existing flow.

Closes https://github.com/netlify/pod-compute/issues/206

### DIFF
--- a/node/deno_config.test.ts
+++ b/node/deno_config.test.ts
@@ -40,6 +40,57 @@ test('Throws a type error if the `importMap` contains anything other than a stri
   await cleanup()
 })
 
+test('Excludes unsupported properties', async () => {
+  const { cleanup, path } = await tmp.dir({ unsafeCleanup: true })
+  const configPath = join(path, 'deno.json')
+  const data = JSON.stringify({
+    compilerOptions: {
+      allowJs: true,
+      lib: ['deno.window'],
+      strict: true,
+    },
+    importMap: 'import_map.json',
+    lint: {
+      files: {
+        include: ['src/'],
+        exclude: ['src/testdata/'],
+      },
+      rules: {
+        tags: ['recommended'],
+        include: ['ban-untagged-todo'],
+        exclude: ['no-unused-vars'],
+      },
+    },
+    fmt: {
+      files: {
+        include: ['src/'],
+        exclude: ['src/testdata/'],
+      },
+      options: {
+        useTabs: true,
+        lineWidth: 80,
+        indentWidth: 4,
+        singleQuote: true,
+        proseWrap: 'preserve',
+      },
+    },
+    test: {
+      files: {
+        include: ['src/'],
+        exclude: ['src/testdata/'],
+      },
+    },
+  })
+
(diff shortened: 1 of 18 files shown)


ascorbic:
Awesome!

node/deno_config.test.ts
```diff
@@ -0,0 +1,57 @@
...
+  const { path } = await tmp.dir()
+  const configPath = join(path, 'deno.json')
+
+  await fs.writeFile(configPath, '{')
+
+  const config = await getConfig(path)
+
+  expect(config).toEqual({})
+
+  await fs.rm(path, { recursive: true })
```
Skn0tt, line 28: according [to the docs](https://www.npmjs.com/package/tmp-promise), `tmp.dir()` also returns a `cleanup` function / a `withDir` helper that does the cleanup for us. That'd remove one line of "noise", and could make the tests easier to read.

node/deno_config.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (config !== undefined) {
+      return normalizeConfig(config, basePath)
+    }
+  }
+}
+
+const getConfigFromFile = async (filePath: string) => {
+  try {
+    const data = await fs.readFile(filePath, 'utf8')
+    const config = parseJSONC(data) as DenoConfigFile
```
Skn0tt, line 32: should we do some minimal validation on the JSON?

Skn0tt (dismissed):
this is rather big, so not sure if I caught everything - but what i've seen, I like!

node/deno_config.test.ts
eduardoboucas: Done in https://github.com/netlify/edge-bundler/pull/256/commits/23bff755512d7dfbc3cb7d07727345ecb04db4dd.

eduardoboucas: I wasn't using it because it needs the `unsafeCleanup` parameter when creating, which I didn't know about. I'll refactor the other tests in a follow-up.

node/deno_config.ts
eduardoboucas: Done in https://github.com/netlify/edge-bundler/pull/256/commits/0c8d8477c620e242b5366513018fe5f3613b43e8.

node/deno_config.test.ts
Skn0tt: Oh, didn't see that in the docs either!

Skn0tt (approved):

eduardoboucas:
Changed to `feat!` because of https://github.com/netlify/edge-bundler/pull/256/files#diff-b2071031d8e70c6ab70129f237c1b91fb71e105d6a0e5e1392091856449055d1R114.

PR #267, opened 2022-12-19
chore: add test

eduardoboucas:

Adds a test asserting that `importMapPaths` can take a path that doesn't correspond to any valid import map file.

### DIFF
--- a/node/bundler.test.ts
+++ b/node/bundler.test.ts
@@ -342,3 +342,25 @@ test('Loads declarations and import maps from the deploy configuration', async (
 
   await cleanup()
 })
+
+test("Ignores entries in `importMapPaths` that don't point to an existing import map file", async () => {
+  const { basePath, cleanup, distPath } = await useFixture('with_import_maps')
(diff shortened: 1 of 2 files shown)


node/bundler.test.ts
```diff
@@ -342,3 +342,25 @@ test('Loads declarations and import maps from the deploy configuration', async (
...
+      function: 'func1',
+      path: '/func1',
+    },
+  ]
+  const result = await bundle([sourceDirectory], distPath, declarations, {
+    basePath,
+    configPath: join(sourceDirectory, 'config.json'),
+    importMapPaths: [join(distPath, 'some-file-that-does-not-exist.json')],
+  })
+  const generatedFiles = await fs.readdir(distPath)
```
Skn0tt, line 360: should we log about the file being ignored, and assert on the log here?

Skn0tt (approved):
looks good, left a small comment :)

Skn0tt (approved):

khendrikse (approved):

PR #268, opened 2022-12-19
fix: read other import map paths when one doesn't exist

eduardoboucas:

Fixes a bug in #265 that prevents additional import map paths from being read when one points to a non-existing file.

### DIFF
--- a/node/import_map.ts
+++ b/node/import_map.ts
@@ -124,7 +124,7 @@ class ImportMap {
   async addFiles(paths: (string | undefined)[], logger: Logger) {
     for (const path of paths) {
       if (path === undefined) {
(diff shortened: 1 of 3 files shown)


node/import_map.ts
```diff
@@ -124,7 +124,7 @@ class ImportMap {
   async addFiles(paths: (string | undefined)[], logger: Logger) {
     for (const path of paths) {
       if (path === undefined) {
-        return
+        continue
```
eduardoboucas, line 127: 🙈

khendrikse (dismissed):
til about continue

khendrikse (approved):

PR #273, opened 2022-12-21
feat!: make `config` export a plain object

eduardoboucas:

Changes the `config` export to expect a plain object instead of a function that returns an object.

https://github.com/netlify/pod-compute/issues/323#issuecomment-1360564040


Skn0tt (approved):

Skn0tt:
looks like some test is still failing

eduardoboucas:
Ah, because the `Config` type in https://edge.netlify.com needs to change. I'll create a PR for that.

PR #274, opened 2022-12-28
feat: move excluded_patterns into function_config

Skn0tt:

Accompanies https://github.com/netlify/stargate/pull/1343. We found that having `excluded_patterns` inside of `routes` doesn't cut it, so we're moving it into a new `function_config` field.

- https://github.com/netlify/stargate/pull/1343
- https://github.com/netlify/pillar-runtime/issues/615
- https://github.com/netlify/netlify-server/pull/2557

### DIFF
--- a/node/bundler.test.ts
+++ b/node/bundler.test.ts
@@ -359,14 +359,14 @@ test('Loads declarations and import maps from the deploy configuration', async (
 
   const manifestFile = await fs.readFile(resolve(distPath, 'manifest.json'), 'utf8')
   const manifest = JSON.parse(manifestFile)
-  const { bundles, routes } = manifest
+  const { bundles, function_config: functionConfig } = manifest
 
   expect(bundles.length).toBe(1)
   expect(bundles[0].format).toBe('eszip2')
   expect(generatedFiles.includes(bundles[0].asset)).toBe(true)
 
   // respects excludedPath from deploy config
-  expect(routes[1].excluded_pattern).toEqual('^/func2/skip/?$')
+  expect(functionConfig.func2).toEqual({ excluded_patterns: ['^/func2/skip/?$'] })
 
   await cleanup()
 })
--- a/node/declaration.ts
+++ b/node/declaration.ts
@@ -44,13 +44,13 @@ export const getDeclarationsFromConfig = (
       const paths = Array.isArray(config.path) ? config.path : [config.path]
 
       paths.forEach((path) => {
-        declarations.push({ ...declaration, ...config, path })
+        declarations.push({ ...declaration, cache: config.cache, path })
       })
 
       // With an in-source config without a path, add the config to the declaration
     } else {
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      const { path, ...rest } = config
+      const { path, excludedPath, ...rest } = config
       declarations.push({ ...declaration, ...rest })
     }
 
@@ -60,15 +60,14 @@ export const getDeclarationsFromConfig = (
   // Finally, we must create declarations for functions that are not declared
   // in the TOML at all.
   for (const name in functionsConfig) {
-    const { ...config } = functionsConfig[name]
-    const { path } = functionsConfig[name]
+    const { cache, path } = functionsConfig[name]
 
     // If we have path specified create a declaration for each path
     if (!functionsVisited.has(name) && path) {
       const paths = Array.isArray(path) ? path : [path]
 
       paths.forEach((singlePath) => {
-        declarations.push({ ...config, function: name, path: singlePath })
+        declarations.push({ cache, function: name, path: singlePath })
       })
     }
   }
--- a/node/manifest.ts
+++ b/node/manifest.ts
@@ -4,20 +4,12 @@ import { join } from 'path'
 import globToRegExp from 'glob-to-regexp'
 
 import type { Bundle } from './bundle.js'
-import { Cache } from './config.js'
+import { Cache, FunctionConfig } from './config.js'
 import type { Declaration } from './declaration.js'
 import { EdgeFunction } from './edge_function.js'
 import { Layer } from './layer.js'
 import { getPackageVersion } from './package_json.js'
 import { nonNullable } from './utils/non_nullable.js'
-
-interface GenerateManifestOptions {
-  bundles?: Bundle[]
-  declarations?: Declaration[]
-  functions: EdgeFunction[]
-  importMap?: string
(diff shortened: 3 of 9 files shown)


node/bundler.test.ts
```diff
@@ -359,14 +359,14 @@ test('Loads declarations and import maps from the deploy configuration', async (
 
   const manifestFile = await fs.readFile(resolve(distPath, 'manifest.json'), 'utf8')
   const manifest = JSON.parse(manifestFile)
-  const { bundles, routes } = manifest
+  const { bundles, function_config: functionConfig } = manifest
```
eduardoboucas, line 362: Could this be just `config` rather than `function_config`?

node/validation/manifest/schema.ts
```diff
@@ -24,6 +24,22 @@ const routesSchema = {
   additionalProperties: false,
 }
 
+const functionConfigSchema = {
+  type: 'object',
+  required: ['excluded_patterns'],
```
eduardoboucas, line 29: Why is it required?

node/bundler.test.ts
Skn0tt: The idea behind calling it `function_config` is that it contains function-specific config, as opposed to "general" config that applies to the full bundle. I fear that `config` is a bit too generic. What do you think?

node/validation/manifest/schema.ts
Skn0tt: 🤷 I don't remember. Removed it in https://github.com/netlify/edge-bundler/pull/274/commits/0fa20a9faee3ed274554d47db8cf298c2f8bbd3a.

node/manifest.ts
```diff
@@ -48,11 +53,23 @@ const generateManifest = ({
   bundles = [],
   declarations = [],
   functions,
+  functionConfig = {},
   importMap,
   layers = [],
 }: GenerateManifestOptions) => {
   const preCacheRoutes: Route[] = []
   const postCacheRoutes: Route[] = []
+  const manifestFunctionConfig: Manifest['function_config'] = Object.fromEntries(
```
eduardoboucas, line 62: I think this means that there will be an entry in `function_config` for every function, even if the object is empty. Is that intended? Could we simply omit a function from the object if it doesn't have any configuration properties?

Skn0tt: Thought the same when I wrote this first. Then I replaced it with https://github.com/netlify/edge-bundler/pull/274/commits/7c497533073b937262212acc4bf423685f1df562, found that to be too much indirection, and went back. Committed it again, let me know what you think ^^

node/declaration.ts
```diff
@@ -44,13 +44,13 @@ export const getDeclarationsFromConfig = (
       const paths = Array.isArray(config.path) ? config.path : [config.path]
 
       paths.forEach((path) => {
-        declarations.push({ ...declaration, ...config, path })
+        declarations.push({ ...declaration, cache: config.cache, path })
```
eduardoboucas, line 47: Not sure I understand this bit. Why are we suddenly setting the `cache` property specifically rather than all of `config`? And does it mean that if we add more configuration properties to ISC, will we have to add them here in addition to `config.ts`?

node/manifest.ts
```diff
@@ -48,11 +53,26 @@ const generateManifest = ({
...
   functions,
+  functionConfig = {},
   importMap,
   layers = [],
 }: GenerateManifestOptions) => {
   const preCacheRoutes: Route[] = []
   const postCacheRoutes: Route[] = []
+  const manifestFunctionConfig: Manifest['function_config'] = {}
+
+  const getFunctionConfig = (name: string) => {
```
eduardoboucas, line 64: I find this a bit confusing. We have a nested function that mutates an object in the outer scope but its name says it's a getter.

If I understand this correctly, the goal of this function is to prime `manifestFunctionConfig` with blank entries that you can safely mutate? Perhaps it would be nicer to make this function explicitly about that, by both changing the name and removing the return value so that it's explicitly a setter.

Then further below you could access `manifestFunctionConfig` directly rather than the return value of `getFunctionConfig`?

node/declaration.ts
Skn0tt: This is necessary because `config` now not only contains `cache` and `path`, but also `excludedPatterns`, which should not be part of the `declaration`. This means that we'll have to manually add all declaration-specific ISC properties here, yes.

Now that I think about it, should `preCache` / `postCache` be part of functions config as well? Are there cases where the same edge function sometimes should run before, and sometimes after cache? (this is out of scope for this PR, but nevertheless we should think about it)

node/manifest.ts
Skn0tt: Tried this out, and having a function called `primeFunctionConfig` also looked weird and added indirection. I tried a different approach where the config is filled up at the start, and then reduced before saving it: https://github.com/netlify/edge-bundler/pull/274/commits/520a241a0dc85cc6f58231cc973f75a9be5b452c How do you like that?

node/declaration.ts
eduardoboucas: > This is necessary because `config` now not only contains `cache` and `path`, but also `excludedPatterns`, which should not be part of the `declaration`.

I guess you could push to the declaration everything except `excludedPatterns`, which you know shouldn't be there. If we add more properties to ISC we wouldn't need to manually include them here?

Happy to defer to you though.

node/manifest.ts
```diff
@@ -44,15 +49,39 @@ interface Route {
 
 const serializePattern = (regex: RegExp) => regex.source.replace(/\\\//g, '/')
 
+const sanitizeEdgeFunctionConfig = (config: Record<string, EdgeFunctionConfig>): Record<string, EdgeFunctionConfig> => {
+  const newConfig = { ...config }
+
+  for (const [name, functionConfig] of Object.entries(newConfig)) {
+    if (functionConfig.excluded_patterns.length === 0) {
+      delete newConfig[name]
```
eduardoboucas, line 57: Personally not a big fan of using `delete` here. I would rather start with an empty object for `newConfig` and then add the individual properties if `excluded_patterns.length !== 0`.

eduardoboucas (dismissed):

node/manifest.ts
Skn0tt: done in https://github.com/netlify/edge-bundler/pull/274/commits/699a05367ae86b26b3ddf4390a0d103fcad9eba6

eduardoboucas (approved):

PR #284, opened 2023-01-11
fix: replace "items" with "additionalProperties"

Skn0tt:
We're currently printing the warning `strict mode: missing type "array" for keyword "items" at "#/properties/function_config" (strictTypes)`. This PR fixes that. Apparently `items` is restricted to use with `type: "array"`, the equivalent is `additionalProperties: ...` (which is the same as `properties: { ".*": ... }`.


Skn0tt:
`properties` requires a static list of properties, which we don't have. I tried using `properties`, and it results in "schema invalid".

eduardoboucas:
Ah, makes sense!

PR #287, opened 2023-01-12
fix: throw correct error for top-level exceptions

Skn0tt:
Originally documented in https://github.com/netlify/cli/pull/5370, this is a second reproduction.

When there's a top-level exception in an edge function, we're not serving a helpful error message. We should fix this.

### DIFF
--- a/node/formats/javascript.ts
+++ b/node/formats/javascript.ts
@@ -8,7 +8,7 @@ import { deleteAsync } from 'del'
 import { EdgeFunction } from '../edge_function.js'
 import type { FormatFunction } from '../server/server.js'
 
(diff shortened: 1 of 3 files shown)


eduardoboucas:
To do this, I think we'd need to use dynamic imports rather than import statements, so that we could add a `try/catch` around the import of each function. We need to check how well this will work with ESZIP.

Skn0tt:
What if we did https://github.com/netlify/edge-bundler/pull/287/commits/4ff5919759abe2afd809ed1366786d956ac3968c together with https://github.com/netlify/edge-functions-bootstrap/pull/153?

Skn0tt:
https://github.com/netlify/edge-bundler/pull/287/commits/dba8101f798231f0dc3dbf34f056155f753bcf44 passes tests :)

Skn0tt:
moving to the icebox until we can use dynamic imports.

PR #298, opened 2023-02-02
feat: validate pattern declarations

eduardoboucas:

This PR parses into an AST any paths defined as regular expressions and:

1. Throws an error if they contain negative lookaheads, as they aren't supported by the Go engine
2. Converts named captured groups into regular captured groups, since the JavaScript syntax for named captured groups is different than the one used by the Go engine

Supersedes #226.

### DIFF
--- a/node/declaration.ts
+++ b/node/declaration.ts
@@ -78,8 +78,9 @@ export const getDeclarationsFromConfig = (
 }
 
 // Validates and normalizes a pattern so that it's a valid regular expression
-// for the Go regexp engine.
+// in Go, which is the engine used by our edge nodes.
 export const parsePattern = (pattern: string) => {
+  // Escaping forward slashes with back slashes.
   const normalizedPattern = pattern.replace(/\//g, '\\/')
   const regex = regexpAST.transform(`/${normalizedPattern}/`, {
     Assertion(path) {
--- a/node/declaration.ts
+++ b/node/declaration.ts
@@ -86,7 +86,7 @@ export const parsePattern = (pattern: string) => {
     Assertion(path) {
       // Lookaheads are not supported. If we find one, throw an error.
       if (path.node.kind === 'Lookahead') {
-        throw new Error('Regular expressions with negative lookaheads are not supported')
+        throw new Error('Regular expressions with lookaheads are not supported')
       }
     },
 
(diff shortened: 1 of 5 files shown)


eduardoboucas:
Hmm, the failing test makes me think there's something I missed in transforming the patterns. 🤔

node/manifest.ts
```diff
@@ -134,20 +140,32 @@ const pathToRegularExpression = (path: string) => {
...
+        `Could not parse path declaration of function '${declaration.function}': ${(error as Error).message}`,
+      )
+    }
   }
 
   return pathToRegularExpression(declaration.path)
 }
 
 const getExcludedRegularExpression = (declaration: Declaration) => {
   if ('pattern' in declaration && declaration.excludedPattern) {
```
danez, line 161:
```suggestion
  if ('excludedPattern' in declaration && declaration.excludedPattern) {
```

? Not really a change from this PR but seems wit was already wrong?

danez (approved):
Just one problem. We can fix this in a separate PR though

Otherwise LGTM 🤘

node/declaration.ts
```diff
@@ -75,4 +77,35 @@ export const getDeclarationsFromConfig = (
...
+// Validates and normalizes a pattern so that it's a valid regular expression
+// in Go, which is the engine used by our edge nodes.
+export const parsePattern = (pattern: string) => {
+  // Escaping forward slashes with back slashes.
+  const normalizedPattern = pattern.replace(/\//g, '\\/')
+  const regex = regexpAST.transform(`/${normalizedPattern}/`, {
+    Assertion(path) {
+      // Lookaheads are not supported. If we find one, throw an error.
+      if (path.node.kind === 'Lookahead') {
+        throw new Error('Regular expressions with negative lookaheads are not supported')
```
ascorbic, line 89: It's not just negative lookaheads that aren't supported: positive lookaheads aren't either.

eduardoboucas: Ah-ha! The comment was correct but the error message was not. Fixed in https://github.com/netlify/edge-bundler/pull/298/commits/2efa3f819951ae2a7c1ea2ec28bbb3a85cd43145.

node/manifest.ts
eduardoboucas: Good catch! Fixed in https://github.com/netlify/edge-bundler/pull/298/commits/51d1023da226a871668faee74960267ed539fbee.

danez (approved):

PR #312, opened 2023-02-27
feat: populate generator field if edge function is from a config file

khendrikse:
Related to https://github.com/netlify/pod-compute/issues/423

But I wonder if this solution makes sense. Can we make sure this represents autogenerated functions correctly?


eduardoboucas:
> Can we make sure this represents autogenerated functions correctly?

There are two ways to define a generated function:

1. Write a `manifest.json` file in `.netlify/edge-functions`
2. Use in-source configuration to define the function config options from inside the function file

I think your solution will only catch the first case. To catch both cases, I think we'll need to tell Edge Bundler where generated functions are located. This could be a `generatedFunctionsPath` property, similarly to what we have (or had?) in zip-it-and-ship-it.

node/utils/add_generator_if_missing.ts
```diff
@@ -0,0 +1,6 @@
+import type { Declaration } from '../declaration.js'
+
+const addGeneratorIfMissing = (functions?: Declaration[]) =>
```
eduardoboucas, line 3: I'm curious why you decided to move this into its own util file, since it's just extending an object with an additional property?

khendrikse: 🤔 I am used to splitting things up this way, but now that you mention it, it doesn't make sense in this case no :)

node/bundler.ts
```diff
@@ -121,7 +125,11 @@ const bundle = async (
 
   // Creating a final declarations array by combining the TOML file with the
   // deploy configuration API and the in-source configuration.
-  const declarations = getDeclarationsFromConfig(tomlDeclarations, functionsWithConfig, deployConfig)
+  const declarationsFromConfig = getDeclarationsFromConfig(tomlDeclarations, functionsWithConfig, deployConfig)
+
+  // If any declarations are autogenerated and are missing the generator field
+  // add a default string.
+  const declarations = addGeneratorFieldIfMissing(declarationsFromConfig, functions, internalFunctionsPath)
```
eduardoboucas, line 132: nit: I don't see a reason why `addGeneratorFieldIfMissing` takes an array of declarations. I think it would read nicer if it took a single declaration and then the `.map` sat here. This would make it clear that you're just processing the same declarations array.

```ts
const declarations = declarationsFromConfig.map(declaration => addGeneratorFieldIfMissing(...
```

eduardoboucas: Also, if `internalFunctionsPath` is `undefined` we probably shouldn't even call `addGeneratorFieldIfMissing`?

node/bundler.ts
```diff
@@ -167,5 +175,22 @@ const getBasePath = (sourceDirectories: string[], inputBasePath?: string) => {
...
+) =>
+  declarations.map((declaration) => {
+    const fullFuncPath = functions?.find((func) => func.name === declaration.function)?.path
+
+    // If function path is in the internalFunctionsPath, we assume it is autogenerated.
+    const isInternal = Boolean(
+      internalFunctionsPath && fullFuncPath && isPathInside(fullFuncPath, internalFunctionsPath),
+    )
+
+    const generatorFallback = isInternal ? 'internalFunc' : undefined
```
eduardoboucas, line 191: Curious why you're going with "internal" here if we're saying that these are "generated" functions?

eduardoboucas: Also, if the property is "generator" and the value is "internalFUNC", does it mean that the function generates itself? 😅 

Would it make sense to use something like "netlify"?

khendrikse: I've thought about this. So I wanted to keep it similar to what I did in zip-it-and-ship-it. And essentially when prepopulating, we don't know if it is netlify. we also don't know if it was a plugin or integration. We only know that it was placed in an internal folder. It's a bit icky though, but that is the thought behind it I guess 🤔 I _could_ change it to `plugin/integration` or `netlify`

node/bundler.ts
khendrikse: I love this change, implementing!

eduardoboucas (approved):

PR #325, opened 2023-03-06
chore: cleanup

danez:
This does some devtooling cleanups:

- use `fs.rm` instead of `fs.rmdir` because `fs.rmdir` logs a deprecation warning.
- lock file maintenance, all dependencies in lock-file are updated to latest valid version
- update eslint config, to fix unsupported typescript warning when eslint runs.
- Instead of `FORCE_COLOR=0` environment variable use `chalk.level=0` so that the tests work when not run through npm
- Remove ava rule in renovate config, as we do not use ava here anymore.
- Increase test timeout to 30s


khendrikse (approved):
I feel like you're the king of maintenance ✨

danez:
> I feel like you're the king of maintenance ✨

I ❤️ maintenance.

PR #327, opened 2023-03-07
fix: throw errors when function or isc-config cannot be loaded

danez:
This changes the behaviour and instead of logging a message we now throw during bundling if:

- The edge function file contains a syntax/runtime error
- The config export is not an object
- The config object could not be serialized

Fixes netlify/pod-compute#422
Fixes netlify/pod-compute#426


khendrikse (dismissed):
Thanks for doing this :) I think it's going to make things a lot more clear for the user.

khendrikse (approved):

eduardoboucas:
As I said yesterday, I would recommend a two-stage approach where we start by just logging these and only then start failing builds. This lets us fine-tune our own logic without disrupting people's workflows.

danez:
> As I said yesterday, I would recommend a two-stage approach where we start by just logging these and only then start failing builds. This lets us fine-tune our own logic without disrupting people's workflows.

Can you elaborate? Because logging is what we do right now.

eduardoboucas:
> Can you elaborate? Because logging is what we do right now.

But we're throwing `BundleError` now, which I don't think we're catching anywhere? Won't that fail the build?

danez:
> > Can you elaborate? Because logging is what we do right now.
> 
> But we're throwing `BundleError` now, which I don't think we're catching anywhere? Won't that fail the build?

Sorry what I meant with `right now` is without this PR. Currently the latest version is already logging all these errors.
And yes with this change we are breaking the build now, which IMHO is better than silently ignoring the error and getting an unpredictable edge-function. 
Although this change should be FF i agree, will add that.

danez:
Added the FF

node/feature_flags.ts
```diff
@@ -1,9 +1,10 @@
-const defaultFlags: Record<string, boolean> = {
+const defaultFlags = {
   edge_functions_fail_unsupported_regex: false,
+  edge_functions_invalid_config_throw: false,
 }
 
 type FeatureFlag = keyof typeof defaultFlags
-type FeatureFlags = Record<FeatureFlag, boolean>
+type FeatureFlags = Record<FeatureFlag | string, boolean>
```
eduardoboucas, line 7: Can you explain this change?

eduardoboucas (approved):

node/config.ts
```diff
@@ -39,7 +40,14 @@ const getConfigExtractor = () => {
...
 }
 
-export const getFunctionConfig = async (func: EdgeFunction, importMap: ImportMap, deno: DenoBridge, log: Logger) => {
+export const getFunctionConfig = async (
+  func: EdgeFunction,
+  importMap: ImportMap,
+  deno: DenoBridge,
+  log: Logger,
+  featureFlags: FeatureFlags,
+  // eslint-disable-next-line max-params
```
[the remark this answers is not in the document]
danez: I'd rather not in this PR, as the FF will be removed again at some point anyway.

node/feature_flags.ts
danez: `FeatureFlag` was of type `string`, because of line 1. I remove the type on `defaultFlags` and because of that the type of `FeatureFlag ` is now `"edge_functions_fail_unsupported_regex"|"edge_functions_invalid_config_throw".

So the the final type for `FeatureFlags` is now Record<"edge_functions_fail_unsupported_regex"|"edge_functions_invalid_config_throw" | string, boolean> which provides autocompletion for the default flags, which wasn't the case before.

eduardoboucas: I'm asking because the idea of the `FeatureFlags` type was that we shouldn't use any flag in the codebase that is not in the `defaultFlags` object (maybe the name `defaultFlags` is misleading).

danez: exactly, now it provides the `defaultFlags` as dropdown, but still allows string.

but i could change it to

```suggestion
type FeatureFlag = keyof typeof defaultFlags | string
type FeatureFlags = Record<FeatureFlag, boolean>
```

Then `FeatureFlag` (singular) also allows any FF. WDYT?

eduardoboucas: > Oh misread your comment. I can remove the `|string` again, then only the flags in `defaultFlags` ar valid.

Yeah, I think that's what we want. Unless you think of a reason to allow arbitrary strings?

danez: Ok, I used partial now, because in some places we feed an empty object, so the flags might be boolean or undefined, but only the flags in `defaultFlags` are allowed to be used:

Screenshot 2023-03-07 at 13 24 40

Screenshot 2023-03-07 at 13 24 16

eduardoboucas (dismissed):
LGTM!

eduardoboucas (approved):

PR #328, opened 2023-03-07
feat: propagate onError config property to manifest

Skn0tt:
Part of https://github.com/netlify/pod-compute/issues/433.

I branched off https://github.com/netlify/edge-bundler/pull/327, because this PR also touches validation. If that takes longer to be merged, I can rebase onto `main`.


node/manifest.ts
```diff
@@ -18,6 +18,7 @@ interface Route {
   name?: string
   pattern: string
   generator?: string
+  on_error?: string
```
Skn0tt, line 21: this is the property name that'll be used in nf-server / stargate - let's get this right on the first attempt. I chose  `on_error`, since that makes the relation to the user-facing feature obvious. But i've also seen `failure mode` being used - what term should we use in code? @eduardoboucas leaving that decision to you.

eduardoboucas: I think `onError` is a more intuitive user-facing name, but I do like "failure mode" when referring to the feature itself (otherwise, do we say that a function has an "on error"?). But I see your point about having a consistent name across the board, in which case I think I'd go with "on error".

What do you think?

Skn0tt: I think it's fine to talk about "failure modes" as an attribute of a deployment, but when talking about *this specific feature*, *this specific config setting*, I like calling it "`onError` setting".
"failure modes" is already an overloaded term - it's a functionality that `onError` enables, but there could be other config settings in the future that contribute to a function's failure mode.
In our domain language, I think we should say that "you can use the `onError` setting to tune your edge function's failure modes". Following that, we should use `on_error` as the internal config value.

Skn0tt:
hmm, that test isn't failing for me locally :thinking:

eduardoboucas:
I think @danez worked on that recently.

danez:
This PR is based on mine (?) and doesn't include the latest changes.

khendrikse:
Sorry I tried to see what would happen if I would change the base branch to the `throw-error` branch 🤔 assumed we wouldn't see Dani's changes then anymore.

khendrikse:
@Skn0tt I think it would be helpful for the review if we have a rebase on dani's branch? :)

khendrikse (changes requested):
I remember when I added the generator field, I also had to add it to the manifest validation schema, [this is the PR I made back then](https://github.com/netlify/edge-bundler/pull/306/files) :) I think this field would also have to be added.

Skn0tt:
rebased! I think there's a flag in a test that needs to be removed, but other than that things look good. @danez sorry for working off your branch, I realize this is putting some pressure on you :/

Skn0tt:
if anybody wants to review, then you can look at https://github.com/netlify/edge-bundler/pull/328/commits/d93075526dcc80499fda00d5632337cc4cd38067 in isolation :) I'll leave some notes on there.

node/config.ts
```diff
@@ -89,17 +98,26 @@ export const getFunctionConfig = async (
...
   } catch {
     handleConfigError(func, ConfigExitCode.UnhandledError, stderr, log, featureFlags)
-
-    return {}
   } finally {
     await collector.cleanup()
   }
+
+  if (!isValidOnError(collectorData.onError)) {
+    throw new BundleError(
```
Skn0tt, line 113: since this is new behaviour, I think it's fine to throw on error - no need to use the flag.

node/manifest.ts
```diff
@@ -106,6 +100,7 @@ const generateManifest = ({
       name: declaration.name,
       pattern: serializePattern(pattern),
       generator: declaration.generator,
+      on_error: declaration.onError ?? functionConfig[func.name]?.onError,
```
Skn0tt, line 103: the order here is important: failure modes can be specified on a declaration level, allowing the same function to have different failure modes on different paths. this takes precedence over settings defined in ISC.

node/validation/manifest/index.ts
```diff
@@ -1,6 +1,6 @@
 import Ajv from 'ajv'
-import type { ValidateFunction } from 'ajv/dist/core.js'
 import ajvErrors from 'ajv-errors'
+import type { ValidateFunction } from 'ajv/dist/core.js'
```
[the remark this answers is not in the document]
Skn0tt: and now it failed in CI, sigh [7db6740](https://github.com/netlify/edge-bundler/pull/328/commits/7db6740d620e71470fb561d25cd25957006fedf7)

node/manifest.ts
eduardoboucas: Hmm I think this might be too confusing. Can we just make this configurable on a per-function basis?

khendrikse (approved):
Looks good! I see Eduardo [had a comment](https://github.com/netlify/edge-bundler/pull/328/files#r1128257200) that is worth looking into, but it's a 👍 from me.

khendrikse (changes requested):
Ugh I commented with the wrong status.

node/config.ts
```diff
@@ -27,10 +27,19 @@ export const enum Cache {
   Manual = 'manual',
 }
 
+export type OnError = 'fail' | 'bypass' | `/${string}`
+
+export const isValidOnError = (value: unknown): value is OnError => {
+  if (typeof value === 'undefined') return true
```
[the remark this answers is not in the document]
Skn0tt: age-old insecurities about how JS handles `null` vs `undefined`. blame brendan eich

node/manifest.ts
Skn0tt: okidoke 👍 i'll move it to the functions config object then

Skn0tt: done in https://github.com/netlify/edge-bundler/pull/328/commits/a704875145029f00746734c7eac9c22a72c13c83

node/config.ts
```diff
@@ -89,17 +98,26 @@ export const getFunctionConfig = async (
...
-
-    return {}
   } finally {
     await collector.cleanup()
   }
+
+  if (!isValidOnError(collectorData.onError)) {
+    throw new BundleError(
+      new Error(
+        `The 'onError' config in edge function at '${func.path}' must be one of 'fail', 'bypass', or a path starting with '/', got '${collectorData.onError}'. More on the Edge Functions API at https://ntl.fyi/edge-api.`,
```
eduardoboucas, line 115: nitty nit

```suggestion
        `The 'onError' configuration property in edge function at '${func.path}' must be one of 'fail', 'bypass', or a path starting with '/'. Got '${collectorData.onError}'. More on the Edge Functions API at https://ntl.fyi/edge-api.`,
```

eduardoboucas (approved):

eduardoboucas (approved):
🌊 🚢

PR #339, opened 2023-03-14
fix: enforce leading slash in path and pattern

danez:

This enforces a leading slash in the TypeScript types and also allows an array of paths in the config.

Ref: https://github.com/netlify/pod-compute/issues/378

Typescript types enforce a slash, but the are only internal.
The manifest validation now checks that the pattern starts with `^/`.

licensed-image


node/validation/manifest/index.test.ts
```diff
@@ -97,32 +97,57 @@ describe('bundle', () => {
 })
 
 describe('route', () => {
+  let freshValidateManifest: typeof validateManifest
```
eduardoboucas, line 100: We'll be able to get rid of this once we remove the feature flag, right?

node/validation/manifest/schema.ts
```diff
@@ -18,7 +18,7 @@ const routesSchema = {
       type: 'string',
       format: 'regexPattern',
       errorMessage:
-        'pattern needs to be a regex that starts with ^ and ends with $ without any additional slashes before and afterwards',
+        'pattern needs to be a regex that starts with ^ followed by / and ends with $ without any additional slashes before and afterwards',
```
eduardoboucas, line 21: I think these error messages need some love (perhaps we can ask @klavavej for some advice). But that can go in a follow-up PR.

eduardoboucas (approved):

node/validation/manifest/index.test.ts
danez: Yes this is only because the code only initializes ajv once, but with the FF changing during the tests we have to reset the modules.

node/validation/manifest/schema.ts
klavavej: Thanks for the ping!

Is it possible to serve different error messages for issues in `netlify.toml` vs issues in `manifest.json`? I ask because ideally I'd like to provide a full example of the syntax for each to be super clear about the required format.

Error for issue in `netlify.toml`:

> pattern must be a regex that starts with ^/ and ends with $ for example pattern = "^/blog/[\d]{4}$"

Error for issue in `manifest.json`:

> pattern must be a regex that starts with ^/ and ends with $ for example "pattern": "^/blog/[\d]{4}$"

If that's not possible, we can use an example that focuses on just the value

> pattern must be a regex that starts with ^/ and ends with $ for example ^/blog/[\d]{4}$

node/validation/manifest/schema.ts
```diff
@@ -35,7 +35,7 @@ const functionConfigSchema = {
         type: 'string',
         format: 'regexPattern',
         errorMessage:
-          'excluded_patterns needs to be an array of regex that starts with ^ and ends with $ without any additional slashes before and afterwards',
+          'excluded_patterns needs to be an array of regex that starts with ^ followed by / and ends with $ without any additional slashes before and afterwards',
```
klavavej, line 38: some questions before I can help refine the copy for this error message

1. Is `excluded_patterns` an internal syntax or is this the property that users and frameworks should be using? I ask because the [excluded patterns demo](https://github.com/netlify/excluded-patterns-demo) that I referenced while [documenting exclusions](https://github.com/netlify/docs/pull/2943) uses the syntax `excludedPattern` (camel case singular vs snake case plural)

2. Are customers/frameworks able to enter an array of regexes for excludedPattern or is the support for arrays here an internal implementation detail (maybe for handling multiple declarations for the same function)? If customers/frameworks can enter an array of regexes what is the syntax for doing that in `netlify.toml` and `manifest.json`? I ask because the [excluded patterns demo](https://github.com/netlify/excluded-patterns-demo) that I referenced while [documenting exclusions](https://github.com/netlify/docs/pull/2943) didn't indicate support for arrays. Here's a table of my understanding of what is supported for a single declaration in the various formats (this understanding is what my open docs PR is based on https://github.com/netlify/docs/pull/2943). Please correct anything that I've got wrong

|   | inline Config  | netlify.toml  | manifest.json  |
|---|---|---|---|
| path  | string or array of strings  | single string  |  single string |
| excludedPath |  string or array of strings |  single string |  single string |
| pattern  | not supported  | single regex  |  single regex |
| excludedPattern  | not supported | single regex  | single regex  |

node/validation/manifest/schema.ts
danez: I start with one message for all, and we can look later if we can validate the path from the user config separately.

PR #347, opened 2023-03-21
feat: split user and internal ISC function configs

danez:
Ref: https://github.com/netlify/pod-compute/issues/442

- add a new test for the order of internal functions
- ensures that the cache property is only added to the manifest if not undefined

This does not change the order of functions in the manifest. It only splits user and internal ISC functions, so that we can adjust the order to our liking.


node/bundler.ts
```diff
@@ -106,36 +109,40 @@ const bundle = async (
...
-  )
+  const internalFunctionsWithConfig: Record<string, FunctionConfig> = {}
+  for (const func of internalFunctions) {
+    internalFunctionsWithConfig[func.name] = await getFunctionConfig(func, importMap, deno, logger, featureFlags)
+  }
+
+  const userFunctionsWithConfig: Record<string, FunctionConfig> = {}
+  for (const func of userFunctions) {
+    userFunctionsWithConfig[func.name] = await getFunctionConfig(func, importMap, deno, logger, featureFlags)
+  }
```
danez, line 121: This removes the parallelism of `getFunctionConfig` but is for now much easier to read, we can try to make `getFunctionConfig`s run in parallel later again.

khendrikse:
We are already passing `internalSrcFolder` to `bundle` :) https://github.com/netlify/edge-bundler/blob/a3b01c08c5bcb734d7b66eedcdf45598abd87640/node/bundler.ts#L35

danez:
> We are already passing `internalSrcFolder` to `bundle` :)
> 
> https://github.com/netlify/edge-bundler/blob/a3b01c08c5bcb734d7b66eedcdf45598abd87640/node/bundler.ts#L35

Yes I removed that and instead made it an param, because it is now not optional anymore.

node/bundler.ts
```diff
@@ -106,36 +109,40 @@ const bundle = async (
...
+  )
 
   const manifest = await writeManifest({
     bundles: [functionBundle],
     declarations,
     distDirectory,
     featureFlags,
     functions,
-    functionConfig: functionsWithConfig,
+    userFunctionConfig: userFunctionsWithConfig,
```
khendrikse, line 144: I'm curious why we don't want to pass it along as a single object. Right now it is merged together in the `generateManifest` function anyway. It can become super useful to pass them separately for future purposes, but I don't see we have to do it now already?

khendrikse (dismissed):
Had a comment, but feel free to ignore it if you don't see it as important enough to change

node/bundler.ts
eduardoboucas: `getFunctionConfig` means spinning up the Deno CLI, running a function, writing the result to a file. This is a non-trivial operation. Have we considered the impact of losing parallelisation here?

node/bundler.ts
```diff
@@ -32,11 +32,11 @@ interface BundleOptions {
   onAfterDownload?: OnAfterDownloadHook
   onBeforeDownload?: OnBeforeDownloadHook
   systemLogger?: LogFunction
-  internalSrcFolder?: string
 }
 
 const bundle = async (
   sourceDirectories: string[],
+  internalSourceDirectory: string,
```
eduardoboucas, line 39: I'm not sure why you're moving this parameter and introducing a breaking change in the process. Is it because you want to make it mandatory? If so, why? If we don't receive an internal source directory, we can just consider all functions to be user-generated? Or am I missing something?

node/declaration.ts
```diff
@@ -59,15 +60,20 @@ export const mergeDeclarations = (
...
 
     // If we have a path specified, create a declaration for each path.
     if (!functionsVisited.has(name) && path) {
       const paths = Array.isArray(path) ? path : [path]
 
       paths.forEach((singlePath) => {
-        declarations.push({ cache, function: name, path: singlePath })
+        const declaration: Declaration = { function: name, path: singlePath }
+        // only add cache if not null or undefined
+        if (cache) {
```
eduardoboucas, line 73: Feels like this comment could be encoded into the actual assertion?

```suggestion
        if (cache !== null && cache !== undefined) {
```

node/bundler.ts
danez: Yeah the API is not perfect. the problem i had conceptually was the before this PR we supplied the user and internal source directory in `sourceDirectories` and additionally supplied the same internal path in the options. In order to split the configs we need to know which path from `sourceDirectories` is the internal one.

What I tried by changing the API is making it somehow clear that when one supplies an internal function directory in `sourceDirectories` that one also have to specify the optional `internalSourcePath` option, as otherwise the generated result might be wrong.

So instead I opted for removing the internalSourcePath `option` removing it from `sourceDirectories` and adding it as its own parameter, because `build` sets it anyway always.

I could move the param back to options and make it optional, but I find this a weird API that one of the main inputs is hidden in the options. `edge-bundler` API doesn't seem to be currently built for the use case of differentiating between directories that have a different type.

One other option I thought about was to supply `sourceDirectories` not as `string[]` but instead `{ dir: string, type: "internal"| "user" }[]` or similar.

Not sure what would be better.

eduardoboucas: I see what you mean. For what it's worth, we use that [same concept in zip-it-and-ship-it](https://github.com/netlify/zip-it-and-ship-it#internalsrcfolder). The initial idea was to keep these packages as dumb as possible, but that's no longer the case as Edge Bundler needs to know how to order the declarations.

I still think the current API works as long as we use it properly, and keeping it saves you the hassle of handling breaking changes.

But I'm happy to defer to you. Just wanted to clarify this and ensure you were aware that the data is already being passed in.

node/bundler.ts
danez: Okay it is parallel again

node/bundler.ts
```diff
@@ -62,7 +62,7 @@ const bundle = async (
     onAfterDownload,
     onBeforeDownload,
   }
-  const internalFunctionsPath = internalSrcFolder && resolve(internalSrcFolder)
```
danez, line 65: build does already send an absolute path so not sure why we would need this.

khendrikse: I think it was more as a failsafe?

khendrikse (approved):

PR #348, opened 2023-03-21
feat: move non-route related ef configs to function_config in manifest

khendrikse:
…unction_config

Example: This fixes #5012

**Checklist**


node/bundler.test.ts
```diff
@@ -321,7 +321,7 @@ test('Processes a function that imports a custom layer', async () => {
   await cleanup()
 })
 
-test('Loads declarations and import maps from the deploy configuration', async () => {
+test('Loads declarations and import maps from the deploy configuration and in-source-config', async () => {
```
eduardoboucas, line 324:
```suggestion
test('Loads declarations and import maps from the deploy configuration and in-source config', async () => {
```

node/bundler.ts
```diff
@@ -173,18 +172,31 @@ const getBasePath = (sourceDirectories: string[], inputBasePath?: string) => {
...
-) => {
-  const fullFuncPath = functions?.find((func) => func.name === declaration.function)?.path
-
+const addMissingConfigFields = ({
+  name,
+  config,
+  functions,
+  declarations,
+  internalFunctionsPath,
+}: {
```
eduardoboucas, line 181: nit: Could we move these options into an interface?

node/bundler.ts
```diff
@@ -173,18 +172,31 @@ const getBasePath = (sourceDirectories: string[], inputBasePath?: string) => {
...
+  name: string
+  config: FunctionConfig
+  functions: EdgeFunction[]
+  declarations: Declaration[]
+  internalFunctionsPath?: string
+}) => {
+  const fullFuncPath = functions?.find((func) => func.name === name)?.path
+  // Check any declarations deprecated name and generator fields
+  const declaration = declarations?.find((decl) => decl.function === name)
```
eduardoboucas, line 190:
```suggestion
  const declaration = declarations?.find((decl) => decl.function === name)

```

node/bundler.ts
```diff
@@ -173,18 +172,31 @@ const getBasePath = (sourceDirectories: string[], inputBasePath?: string) => {
...
 }
 
-export const addGeneratorFieldIfMissing = (
-  declaration: Declaration,
-  functions: EdgeFunction[],
-  internalFunctionsPath?: string,
-) => {
-  const fullFuncPath = functions?.find((func) => func.name === declaration.function)?.path
-
+const addMissingConfigFields = ({
```
eduardoboucas, line 175: I don't think "missing config fields" is very descriptive — what's missing and why? Perhaps we can add a comment explaining why we need this function?

node/manifest.ts
```diff
@@ -43,6 +43,17 @@ interface GenerateManifestOptions {
...
 
+const removeEmptyConfigValues = (functionConfig: EdgeFunctionConfig) =>
+  Object.entries(functionConfig).reduce((acc, [key, value]) => {
+    if (value && !(Array.isArray(value) && value.length === 0)) {
+      return { ...acc, [key]: value }
+    }
+    return acc
+  }, {} as EdgeFunctionConfig)
+
+const hasAnyConfigValues = (functionConfig: EdgeFunctionConfig) =>
```
eduardoboucas, line 54: I don't love the fact that we're defining above what a config object is and then having another allow-list here, so adding a new property will mean adding it in two places. I can see people getting tripped up by that.

I don't have an obvious alternative to offer without getting into more complex schema validation, so I'm happy to defer to you.

node/manifest.ts
```diff
@@ -43,6 +43,17 @@ interface GenerateManifestOptions {
   layers?: Layer[]
 }
 
+const removeEmptyConfigValues = (functionConfig: EdgeFunctionConfig) =>
```
eduardoboucas, line 46: Why do we need to remove empty values?

eduardoboucas (commented):
Left a couple of questions before doing a second pass.

node/bundler.test.ts
khendrikse: ✅ taking it along in my commit!

node/bundler.ts
```diff
@@ -173,18 +172,31 @@ const getBasePath = (sourceDirectories: string[], inputBasePath?: string) => {
...
+  internalFunctionsPath,
+}: {
+  name: string
+  config: FunctionConfig
+  functions: EdgeFunction[]
+  declarations: Declaration[]
+  internalFunctionsPath?: string
+}) => {
+  const fullFuncPath = functions?.find((func) => func.name === name)?.path
+  // Check any declarations deprecated name and generator fields
```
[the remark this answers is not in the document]
khendrikse: ✅ taking it along in my commit!

node/bundler.ts
khendrikse: ✅ taking it along in my commit!

node/declaration.ts
```diff
@@ -5,6 +5,7 @@ import { FunctionConfig, Path } from './config.js'
 interface BaseDeclaration {
   cache?: string
   function: string
+  // todo: remove these two after a while and only support in-source-config for non-route related configs
```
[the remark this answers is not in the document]
khendrikse: ✅ taking it along in my commit!

node/bundler.ts
khendrikse: I shall!

node/bundler.ts
khendrikse: I decided to switch it up a bit and make it, hopefully, more descriptive, and give functions a clear single task.

node/manifest.ts
khendrikse: I tried it this way because [on this line](https://github.com/netlify/edge-bundler/pull/348/files#diff-e659ded745234fd55b239ea6728eef800523583941fd1542d88b44373cd39e8aL56) we were checking some of the values. 

What I could do, is allow for anything in the config to be passed to the function_config. This would mean that we get anything defined in it. Including `cache: manual` for example. Then we only need to filter out any config that is empty.

node/manifest.ts
khendrikse: I did this because the logic before this would allow for undefined values to be passed. For example. if we had `on_error` but `functionConfig.excluded_patterns = []` we would otherwise still get that empty array added, or fields that were undefined. Like `generator: undefined`. Essentially I tried to optimise [this line](https://github.com/netlify/edge-bundler/pull/348/files#diff-e659ded745234fd55b239ea6728eef800523583941fd1542d88b44373cd39e8aL56) so I wouldn't have to add a bunch of other config values. And even if I did, if one of them was true, it would add undefined values for the others.

danez:
So we are not allowing `generator` and `name` being set through the manifest? Why is that?

danez:
FYI: I'm also working on the same code for splitting ISC up into internal vs user. https://github.com/netlify/edge-bundler/pull/347/files

khendrikse:
> So we are not allowing `generator` and `name` being set through the manifest? Why is that?

@danez the idea is that anything that is declared in the declarations for the manifest is mostly meant for routing information. ISC is for configuration on a function level.

khendrikse:
> FYI: I'm also working on the same code for splitting ISC up into internal vs user. https://github.com/netlify/edge-bundler/pull/347/files

Let's first get that merged then!

node/bundler.ts
```diff
@@ -186,19 +183,42 @@ const getBasePath = (sourceDirectories: string[], inputBasePath?: string) => {
...
-  const isInternal = Boolean(internalFunctionsPath && fullFuncPath && isPathInside(fullFuncPath, internalFunctionsPath))
+// Merged any old-style non-in-source configs from declaration.
+const mergeWithDeclarationConfig = ({ name, config, declarations }: MergeWithDeclarationConfigOptions) => {
+  const declaration = declarations?.find((decl) => decl.function === name)
 
-  const generatorFallback = isInternal ? 'internalFunc' : undefined
-  return { ...declaration, generator: declaration.generator || generatorFallback }
+  return {
+    ...config,
+    name: declaration?.name || config.name,
```
danez, line 198: This slightly confused me, until i figured that this is the displayName and the passed in name is the functionName. Could we make this more clear? maybe alias the input `name` as `functionName` or add a comment here that this is the display name

node/manifest.ts
```diff
@@ -54,14 +65,29 @@ const sanitizeEdgeFunctionConfig = (config: Record<string, EdgeFunctionConfig>):
   const newConfig: Record<string, EdgeFunctionConfig> = {}
 
   for (const [name, functionConfig] of Object.entries(config)) {
-    if (functionConfig.excluded_patterns.length !== 0 || functionConfig.on_error) {
-      newConfig[name] = functionConfig
+    const newFunctionConfig = removeEmptyConfigValues(functionConfig)
+
+    if (hasAnyConfigValues(newFunctionConfig)) {
```
danez, line 70: Wouldn't this be the same as `Object.keys(newFunctionConfig).length > 0`?

node/manifest.ts
```diff
@@ -44,6 +44,17 @@ interface GenerateManifestOptions {
   userFunctionConfig?: Record<string, FunctionConfig>
 }
 
+const removeEmptyConfigValues = (functionConfig: EdgeFunctionConfig) =>
+  Object.entries(functionConfig).reduce((acc, [key, value]) => {
+    if (value && !(Array.isArray(value) && value.length === 0)) {
+      return { ...acc, [key]: value }
+    }
+    return acc
+  }, {} as EdgeFunctionConfig)
```
danez, line 53:
```suggestion
  Object.fromEntries(Object.entries(functionConfig).filter(([_, value]) => value && !(Array.isArray(value) && value.length === 0)));
```

this is totally subjective, but I'm not a big fan of reduce, unless it is really reducing something. So I would find this much more readable and probably less objects created, but up to you.

node/bundler.ts
danez: so we could do 
```suggestion
    name,
```
 here?

node/manifest.ts
khendrikse: I think I would find this less readable because it hides a lot of logic and I need to read it a lot of times to understand what is happening. That's why I prefer the reduce option

node/bundler.ts
khendrikse: I changed the way I named it now

node/manifest.ts
khendrikse: For some reason, before that didn't work... but now it does! Nice catch!

node/bundler.ts
```diff
@@ -186,19 +183,44 @@ const getBasePath = (sourceDirectories: string[], inputBasePath?: string) => {
...
+  ...config,
+  generator: config.generator || 'internalFunc',
+})
+
+interface CreateFunctionConfigOptions {
+  internalFunctionsWithConfig: Record<string, FunctionConfig>
+  declarations: Declaration[]
+}
+
+const createFunctionConfig = ({ internalFunctionsWithConfig, declarations }: CreateFunctionConfigOptions) =>
```
eduardoboucas, line 215: nit: I find this function a bit hard to follow. The name isn't super descriptive — it says it creates function configs but it's iterating one of the parameters and calling two other functions inside. I feel like an abstraction like `addGeneratorFallback`, which is simply adding a field, is adding to the confusion more than the benefits it brings.

node/declaration.ts
```diff
@@ -5,6 +5,7 @@ import { FunctionConfig, Path } from './config.js'
 interface BaseDeclaration {
   cache?: string
   function: string
+  // todo: remove these two after a while and only support in-source config for non-route related configs
```
eduardoboucas, line 8: What's "after a while"? Should this TODO be an issue instead?

node/manifest.ts
```diff
@@ -54,14 +62,29 @@ const sanitizeEdgeFunctionConfig = (config: Record<string, EdgeFunctionConfig>):
...
 
   return newConfig
 }
 
+const addExcludedPatterns = (
+  name: string,
+  manifestFunctionConfig: Record<string, EdgeFunctionConfig>,
+  excludedPath?: Path | Path[],
+) => {
+  if (excludedPath) {
```
eduardoboucas, line 80: nit: You could early-return and avoid nesting all these statements.

```js
if (!excludedPath) {
  return
}
```

PR #359, opened 2023-04-05
fix: remove duplicate functions and let .js take precedence

khendrikse:
Solving: https://github.com/netlify/pod-compute/issues/461

As [documented here](https://docs.netlify.com/edge-functions/get-started/#deploy),

> If a project has TypeScript and JavaScript edge functions with the same name, for example, my-function.ts and my-function.js, the TypeScript function is ignored while the JavaScript function is deployed.

After testing this out, it appears that when deployed, it works the other way around. The typescript function takes precedence over the javascript file.

We need to make sure it is as described in the documentation.

This PR should fix this.

### DIFF
--- a/node/finder.ts
+++ b/node/finder.ts
@@ -8,27 +8,29 @@ import { nonNullable } from './utils/non_nullable.js'
 // with a lower index meaning a higher precedence over the others
 const ALLOWED_EXTENSIONS = ['.js', '.jsx', '.ts', '.tsx']
 
-const removeDuplicatesByExtension = (functions: string[]) => {
+export const removeDuplicatesByExtension = (functions: string[]) => {
   const seen = new Map()
 
-  return functions.reduce((acc, path) => {
-    const { ext, name } = parse(path)
-    const extIndex = [...ALLOWED_EXTENSIONS].indexOf(ext)
+  return Object.values(
+    functions.reduce((acc, path) => {
+      const { ext, name } = parse(path)
+      const extIndex = ALLOWED_EXTENSIONS.indexOf(ext)
 
-    if (!seen.has(name) || seen.get(name) > extIndex) {
-      seen.set(name, extIndex)
-      return [...acc, path]
-    }
+      if (!seen.has(name) || seen.get(name) > extIndex) {
+        seen.set(name, extIndex)
+        return { ...acc, [name]: path }
+      }
 
-    return acc
-  }, [] as string[])
+      return acc
+    }, {}),
+  ) as string[]
 }
 
 const findFunctionInDirectory = async (directory: string): Promise<EdgeFunction | undefined> => {
   const name = basename(directory)
-  const candidatePaths = [...ALLOWED_EXTENSIONS]
-    .flatMap((extension) => [`${name}${extension}`, `index${extension}`])
-    .map((filename) => join(directory, filename))
+  const candidatePaths = ALLOWED_EXTENSIONS.flatMap((extension) => [`${name}${extension}`, `index${extension}`]).map(
+    (filename) => join(directory, filename),
+  )
 
   let functionPath
 
--- a/node/finder.test.ts
+++ b/node/finder.test.ts
@@ -1,0 +1,20 @@
+import { test, expect } from 'vitest'
+
+import { removeDuplicatesByExtension } from './finder.js'
+
+test('filters out any duplicate files based on the extension', () => {
+  const functions = [
+    'file1.js',
+    'file1.ts',
+    'file2.tsx',
+    'file2.jsx',
+    'file3.tsx',
+    'file3.js',
+    'file4.ts',
+    'file5.ts',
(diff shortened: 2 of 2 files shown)


khendrikse:
I'm also going to add a similar logic for .jsx vs .tsx

Skn0tt:
> This IS a breaking change, even though it is specified in the docs that it should have worked like this before. I'm unsure how we should proceed around that.

My 2cts are that if we had docs for *this* behaviour, then it's a fix and not a breaking change

node/finder.ts
```diff
@@ -4,8 +4,26 @@ import { basename, extname, join } from 'path'
 import { EdgeFunction } from './edge_function.js'
 import { nonNullable } from './utils/non_nullable.js'
 
+// the order of the allowed extensions is also the order we remove duplicates
+// with a lower index meaning a higher precedence over the others
 const ALLOWED_EXTENSIONS = new Set(['.js', '.jsx', '.ts', '.tsx'])
```
eduardoboucas, line 9: Sets don't (theoretically) hold order, so if order now matters should this become an array?

node/finder.ts
```diff
@@ -4,8 +4,26 @@ import { basename, extname, join } from 'path'
...
 
+const removeDuplicatesByExtension = (functions: EdgeFunction[]) => {
```
eduardoboucas, line 11: I find this logic a bit cumbersome and solving a problem that we created ourselves in the first place. For example, if we changed the `find*` functions to hold a `Map` of function names to paths and we checked whether a function name already exists before we add it, we would only need to check the extension order when we found a duplicate. I think that would be clearer than iterating all functions again, then iterating all extensions for each function, etc.

Not a blocker, though.

khendrikse: I tried improving this, curious if this is kind of what you meant

eduardoboucas: Not really. My thinking was that if we had a `Map` like the one below that is passed around to `findFunctionInDirectory`:

```
{
  "func1" => "/some/path/func1.js",
  "func2" => "/some/other/path/func2.ts"
}
```

If it sees `func1.ts`, `func2.js` and `func3.ts` in a directory, it could do the following:

- There's already a `func1` and it's JS, which takes precedence, so do nothing;
- There's already a `func2` and it's TS, whereas the one we're evaluating is JS and that takes precedence, so rewrite that key with the JS path;
- There's no `func3`, so write that key with the TS path.

This way you wouldn't need a `removeDuplicatesByExtension` function at all.

khendrikse: We'd then have to pass it to both `findFunctionInPath` and `findFunctionInDirectory`, as `findFunctionInDirectory` is a part of `findFunctionInPath`. 

Would it not just be moving the location where we check the files? 

With the current solution, we don't even need to check anything in `findFunctionInPath` and `findFunctionInDirectory`, because I already filter out the duplicates before we do any other logic. In those other places, it might not need `removeDuplicatesByExtension`, but I imagine it would still need to do similar checks as I'm doing in that function right now.

eduardoboucas: > We'd then have to pass it to both `findFunctionInPath` and `findFunctionInDirectory`, as `findFunctionInDirectory` is a part of `findFunctionInPath`.

Yes, you'd need to pass that map to any methods responsible for finding functions. You'd be passing by reference the bucket where they need to put data, rather than asking them to return it.



> Would it not just be moving the location where we check the files?

I don't think so. Rather than compiling a list of all functions (including deduplicates) and then stripping duplicates from that list, you'd be constructing a list that never has duplicates. So once you're done traversing, there's nothing else you need to do to that list.

It's a different way of approaching the problem, and my suggestion was merely a suggestion.

eduardoboucas: It's also _theoretically_ less efficient, because right now removing duplicates means iterating X functions and checking Y extensions inside each iteration (so X * Y in total). With the other method, you'd check for duplicates as part of the iteration you'll already do on each directory, and checking for the duplicate itself would be cheap (since it's a `Map`).

But we're talking about very small array sizes anyway, so it shouldn't make any practical difference (hence the _theoretically_).

eduardoboucas (approved):

node/finder.ts
khendrikse: > With the other method, you'd check for duplicates as part of the iteration you'll already do on each directory, and checking for the duplicate itself would be cheap (since it's a `Map`).

Hmm this is true. I'd need to tweak your idea a bit though, and instead of having a Map with name => path, I think using part of the current implementation would be more performant, where we have name => extension. This way we don't have to grab the extension from the  the existing items in the Map the whole time. 

I'm going to try it for a bit!

eduardoboucas: Don't feel like you have to, though. You're closer to the implementation than I am at this point, so it's possible I'm missing some nuance and I trust your judgement.

The approval is there, so feel free to ship as is.

PR #360, opened 2023-04-06
fix: add repro for asana case

Skn0tt:
A test case for Asana's usecase as per https://netlify.slack.com/archives/C0183KVDHFH/p1680717149963289?thread_ts=1679497831.390369&cid=C0183KVDHFH.

### DIFF
--- a/node/manifest.test.ts
+++ b/node/manifest.test.ts
@@ -93,6 +93,23 @@ test('Generates a manifest with excluded paths and patterns', () => {
   expect(manifest.bundler_version).toBe(env.npm_package_version as string)
 })
 
+test('TOML-Defined paths can be combined with ISC-defined excluded paths', () => {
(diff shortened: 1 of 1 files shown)


node/manifest.test.ts
```diff
@@ -93,6 +93,23 @@ test('Generates a manifest with excluded paths and patterns', () => {
   expect(manifest.bundler_version).toBe(env.npm_package_version as string)
 })
 
+test('TOML-Defined paths can be combined with ISC-defined excluded paths', () => {
```
eduardoboucas, line 96:
```suggestion
test('TOML-defined paths can be combined with ISC-defined excluded paths', () => {
```

eduardoboucas (approved):

Skn0tt:
the test failure is interesting 🤔

PR #368, opened 2023-04-13
feat: add `excluded_patterns` on route level

Skn0tt:
To properly support TOML-defined `exclude_pattern`, this PR adds a new `excluded_patterns` field that lives within `routes`. This is in-line with differentiating function- and route-level configuration: ISC config applies to the function, so it lives in `function_config`, TOML config applies to a route mapping, so it lives in `routes`.

Part of https://github.com/netlify/pod-compute/issues/471.

### DIFF
--- a/node/config.test.ts
+++ b/node/config.test.ts
@@ -201,6 +201,7 @@ test('Loads function paths from the in-source `config` function', async () => {
     {
       function: 'user-func2',
       path: '/user-func2',
+      excludedPath: '/user-func2/excluded',
     },
   ]
   const result = await bundle([internalDirectory, userDirectory], distPath, declarations, {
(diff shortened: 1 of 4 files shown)


node/manifest.ts
```diff
@@ -111,7 +112,9 @@ const generateManifest = ({
     manifestFunctionConfig[name] = { ...manifestFunctionConfig[name], on_error: onError }
   }
 
-  for (const [name, { excludedPath, path, onError, ...rest }] of Object.entries(internalFunctionConfig)) {
+  for (const [name, { excludedPath, cache: _cache, path: _path, onError, ...rest }] of Object.entries(
```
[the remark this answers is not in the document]
Skn0tt: (not related to the PR topic, but I figured it's small enough to include here)

node/manifest.ts
```diff
@@ -139,7 +142,11 @@ const generateManifest = ({
     )
 
     if (excludedPattern) {
-      manifestFunctionConfig[func.name].excluded_patterns.push(serializePattern(excludedPattern))
+      if (featureFlags?.edge_functions_excluded_patterns_on_route) {
+        route.excluded_patterns = [serializePattern(excludedPattern)]
```
eduardoboucas, line 146: You're always assigning a one-element array here. Will it have multiple when added via ISC?

node/config.test.ts
```diff
@@ -231,10 +232,63 @@ test('Loads function paths from the in-source `config` function', async () => {
...
+    {
+      function: 'user-func2',
+      pattern: '^/user-func2/?$',
+      // excludedPath is defined in TOML, so it's stored in `routes`
+      excluded_patterns: ['^/user-func2/excluded/?$'],
+    },
+    { function: 'framework-func1', pattern: '^/framework-func1/?$' },
+    { function: 'user-func1', pattern: '^/user-func1/?$' },
+    { function: 'user-func3', pattern: '^/user-func3/?$' },
+    // excludedPath is defined in ISC, so it's not here but in in `function_config`
```
eduardoboucas, line 281: If what we find in ISC generates one entry in `routes`, could the excluded path also make its way there?

eduardoboucas (commented):
This ISC vs. TOML / `routes` vs. `function_config` is 🤯

node/manifest.ts
Skn0tt: yup - extracted that into https://github.com/netlify/edge-bundler/pull/369 as a follow-up, since I didn't want to complicate this PR. But I can also add that commit into this PR, probably makes reviewing easier.

node/config.test.ts
Skn0tt: We could do that for clarity, but it'd mean that data is present in both places (we still need it in `function_config` since it applies to all functions). I'd leave it out 🤷‍♂️ 

I think what's a bit weird here is that in the ISC `{ path: "/*", excludedPath: ["/*.css", "/*.js"] }`, `excludedPath` not only applies to `/*`, but also to _all_ other routes defined in the TOML, whereas an `excludedPath` defined in the TOML only applies to its accompanying `path`. That makes the mental model not as straight-forward to explain, but I think it can be expressed in docs something like this:



```md
# Edge Functions Routing

To assign an Edge Function to a specific route,
set the `path` field in your function config to the URL pattern
you want to apply the function to.
You can exclude certain paths, like static assets, using `excludedPath`.
This means that the function will never be run for those assets.

You can use `netlify.toml` to define more routes.
Again, set the `path` field to the URL pattern you want to apply the function to.
If you want to skip subpath, use `excludedPath`.
The excluded paths won't apply to the whole function, but only to that path.
This can be used to skip an edge function for certain subpages.
```

eduardoboucas: It's interesting that you say the exclusions in ISC apply to every route of that function. I could also see us saying that ISC exclusions apply to the paths defined in ISC, and so the exclusions would go to all the routes generated from ISC. If that makes sense?

Skn0tt:
moving to draft until we've finished the discussion in https://github.com/netlify/pod-compute/issues/471

Skn0tt:
this was shipped in a different PR.

PR #376, opened 2023-04-27
fix: front slashes only get escaped now if they aren't already

### LINKED ISSUE #375: Error thrown when escaping front slashes in regex strings that have already been escaped
Thanks for reporting this bug!

Please search other issues to make sure this bug has not already been reported.

Then fill in the sections below.

**Describe the bug**

there’s logic in the edge bundler that escapes front slashes for regexes, but doesn’t take into account if they are already escaped. 

See https://github.com/netlify/edge-bundler/blob/9d798de1bc646e64770c98e8c1720ba13b886e15/node/declaration.ts#L134
So basically

https://github.com/netlify/edge-bundler/blob/9d798de1bc646e64770c98e8c1720ba13b886e15/node/manifest.ts#L183-L197

runs and `parsePattern` converts

`"^(?:\\/(_next\\/data\\/[^/]{1,}))?(?:\\/([^/.]{1,}))\\/api(?:\\/((?:[^\\/#\\?]+?)(?:\\/(?:[^\\/#\\?]+?))*))?(.json)?[\\/#\\?]?$"`

to

`"^(?:\\\\/(_next\\\\/data\\\\/[^\\/]{1,}))?(?:\\\\/([^\\/.]{1,}))\\\\/api(?:\\\\/((?:[^\\\\/#\\?]+?)(?:\\\\/(?:[^\\\\/#\\?]+?))*))?(.json)?[\\\\/#\\?]?$"`

which results in the following error

```
Unexpected token: "/" at 1:7.
◈ Rewrote URL to /static/chunks/pages/shows/static/[id]-52e80ca069ee1610.js
Function 'next_middleware' uses an unsupported regular expression and will not be invoked: 

/^(?:\\/(_next\\/data\\/[^\/]{1,}))?(?:\\/([^\/.]{1,}))?\\/api(?:\\/((?:[^\\/#\?]+?)(?:\\/(?:[^\\/#\?]+?))*))?(.json)?[\\/#\?]?$/ 
```

because there is an extra escaped backslash added.

In our scenario on the frameworks team, the failing regular expressions are typicially generated from the Next.js framework.

The first question I have is do we need to normalize these regex patterns at the edge bundler level?

`const normalizedPattern = pattern.replace(/\//g, '\\/')`

In user and framework land, if they weren't escaped properly, I would expect it to fail if I did not escape things properly. Is there a use case on the edge functions team for this? I would assume so given the code was added. 😎

For more context see https://netlify.slack.com/archives/C01TKAEBP3Q/p1682545091365449

**Configuration**

Please enter the following command in a terminal and copy/paste its output:

```bash
npx envinfo --system --binaries
```

  System:
    OS: macOS 13.3.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 139.58 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.15.0 - ~/.n/bin/node
    Yarn: 1.22.19 - ~/.n/bin/yarn
    npm: 9.5.0 - ~/.n/bin/npm

**Pull requests**

Pull requests are welcome! If you would like to help us fix this bug, please check our
[contributions guidelines](../blob/main/CONTRIBUTING.md).

nickytonline:

Currently Next.js middleware breaks because the edge bundler modifies the regular expression matchers for URLs that Next.js generates.

fixes #375 
fixes https://github.com/netlify/pod-ecosystem-frameworks/issues/414
relates to https://github.com/netlify/next-runtime/pull/1919

I've modified the regex to do a negative lookbehind to ensure we don't escape a front slash if it's already escaped.

N/A but as I mentioned in the issue, do we want the edge bundler modifying regex patterns?

**Checklist**

### DIFF
--- a/node/declaration.test.ts
+++ b/node/declaration.test.ts
@@ -209,7 +209,7 @@ test('does not escape front slashes in a regex pattern if they are already escap
   expect(actual).toEqual(expected)
 })
 
-test('escapes front slashes in a regex pattern', () => {
+test('Escapes front slashes in a regex pattern', () => {
   const regexPattern = '^(?:/(_next/data/[^/]{1,}))?(?:/([^/.]{1,}))/shows(?:/(.*))(.json)?[/#\\?]?$'
   const expected = '^(?:\\/(_next\\/data\\/[^\\/]{1,}))?(?:\\/([^\\/.]{1,}))\\/shows(?:\\/(.*))(.json)?[\\/#\\?]?$'
   const actual = parsePattern(regexPattern)
--- a/node/declaration.test.ts
+++ b/node/declaration.test.ts
@@ -211,7 +211,7 @@ test('does not escape front slashes in a regex pattern if they are already escap
 
 test('escapes front slashes in a regex pattern', () => {
   const regexPattern = '^(?:/(_next/data/[^/]{1,}))?(?:/([^/.]{1,}))/shows(?:/(.*))(.json)?[/#\\?]?$'
-  const expected = '^(?:\\/(_next\\/data\\/[^\\/]{1,}))?(?:\\/([^\\/.]{1,}))\\/shows(?:\\/(.*))(.json)?[/#\\?]?$'
+  const expected = '^(?:\\/(_next\\/data\\/[^\\/]{1,}))?(?:\\/([^\\/.]{1,}))\\/shows(?:\\/(.*))(.json)?[\\/#\\?]?$'
   const actual = parsePattern(regexPattern)
 
   expect(actual).toEqual(expected)
--- a/node/declaration.test.ts
+++ b/node/declaration.test.ts
@@ -209,7 +209,7 @@ test('does not escape front slashes in a regex pattern if they are already escap
   expect(actual).toEqual(expected)
 })
 
-test('escapes front slashes in a regex pattern if they are not already escaped', () => {
+test('escapes front slashes in a regex pattern', () => {
   const regexPattern = '^(?:/(_next/data/[^/]{1,}))?(?:/([^/.]{1,}))/shows(?:/(.*))(.json)?[/#\\?]?$'
   const expected = '^(?:\\/(_next\\/data\\/[^\\/]{1,}))?(?:\\/([^\\/.]{1,}))\\/shows(?:\\/(.*))(.json)?[/#\\?]?$'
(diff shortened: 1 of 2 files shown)


node/declaration.ts
```diff
@@ -130,8 +130,9 @@ const createDeclarationsFromFunctionConfigs = (
 // Validates and normalizes a pattern so that it's a valid regular expression
 // in Go, which is the engine used by our edge nodes.
 export const parsePattern = (pattern: string) => {
-  // Escaping forward slashes with back slashes.
-  const normalizedPattern = pattern.replace(/\//g, '\\/')
+  // only escape front slashes if they haven't been escaped already
```
eduardoboucas, line 133: Tiny nit to make this consistent with the rest of the comments. 🤓 

```suggestion
  // Only escape front slashes if they haven't been escaped already.
```

node/declaration.test.ts
```diff
@@ -191,3 +191,28 @@ test('netlify.toml-defined excludedPath are respected', () => {
...
+test('throws if there are lookaheads in a regex pattern', () => {
+  const regexPattern =
+    '^(?:\\/(_next\\/data\\/[^/]{1,}))?(?:\\/([^/.]{1,}))\\/shows(?:\\/((?!99|88).*))(.json)?[\\/#\\?]?$'
+
+  expect(() => {
+    parsePattern(regexPattern)
+  }).toThrow('Regular expressions with lookaheads are not supported')
+})
+
+test('does not escape front slashes in a regex pattern if they are already escaped', () => {
```
eduardoboucas, line 204:
```suggestion
test('Does not escape front slashes in a regex pattern if they are already escaped', () => {
```

eduardoboucas (commented):
LGTM! I left just tiny style nits.

Do you want to move this out of draft?

node/declaration.test.ts
```diff
@@ -191,3 +191,28 @@ test('netlify.toml-defined excludedPath are respected', () => {
 
   expect(declarations).toEqual(expectedDeclarations)
 })
+
+test('throws if there are lookaheads in a regex pattern', () => {
```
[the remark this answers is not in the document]
nickytonline: Ahh, I must have missed that. I'll remove this test then.

PR #381, opened 2023-05-04
fix: remove feature flag for execution order

danez:
The feature flag was rolled out to everyone over a week ago. Removing it is important because right now deployments from the CLI still use the old behaviour.


node/declaration.ts
```diff
@@ -28,40 +28,25 @@ export const mergeDeclarations = (
   userFunctionsConfig: Record<string, FunctionConfig>,
   internalFunctionsConfig: Record<string, FunctionConfig>,
   deployConfigDeclarations: Declaration[],
-  featureFlags: FeatureFlags = {},
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
eduardoboucas, line 31: What's the goal here?

eduardoboucas (approved):

node/declaration.ts
danez: `mergeDeclarations` is public API (used in the CLI), so removing the param is technically a breaking change. And who knows maybe we are going to need it in the future again?

but can also remove it.

eduardoboucas: Got it, makes sense!

PR #392, opened 2023-05-05
feat: replace `glob-to-regexp` with `URLPattern`

eduardoboucas:

Replaces the `glob-to-regexp` library with the web platform standard `URLPattern` (using a polyfill, since it's not yet part of Node.js).

This should retain the same functionality advertised in our docs while paving the way for https://github.com/netlify/pod-compute/issues/459.


node/utils/urlpattern.ts
```diff
@@ -0,0 +1,7 @@
+import { URLPattern } from 'urlpattern-polyfill'
+
+export class ExtendedURLPattern extends URLPattern {
+  // @ts-expect-error Internal property that the underlying class is using but
+  // not exposing.
+  regexp: Record<string, RegExp>
```
danez, line 6: I don't think this is a good idea. This will most probably not be available once this is available in Node.js natively. Also looking at the docs of how `URLPattern` is supposed to be used, I do not think it matches our usecase of converting a glob to a regex.

danez: The mdn docs mention this library: https://github.com/pillarjs/path-to-regexp maybe we should use this?

danez: https://developer.mozilla.org/en-US/docs/Web/API/URL_Pattern_API

danez: Okay, up to you. I personally feel this is a terrible idea to rely on the internals and implementation details of other packages.

danez (dismissed):

node/utils/urlpattern.ts
Skn0tt: I share the sentiment of this feeling "bad", but I think it's an example of us knowing the rules enough that we're allowed to break them. Since we're using a polyfill, the underlying implementation won't depend on the runtime, and since we have tests we'll notice when a change in the polyfill breaks this. I think it's fine - and if we find problems, as a last resort we can still reimplement the spec ourselves.

node/manifest.test.ts
```diff
@@ -127,7 +126,7 @@ test('Filters out internal in-source configurations in user created functions',
   ]
   const declarations: Declaration[] = [
     { function: 'func-1', path: '/f1/*' },
-    { function: 'func-2', pattern: '^/f2/.*/?$' },
+    { function: 'func-2', pattern: '^/f2(?:/(.*))/?$' },
```
Skn0tt, line 129: I've checked the impact of changing from `/.*` to `(?:/(.*))` with this go benchmark:

```golang
import (
	"regexp"
	"testing"
)

func BenchmarkCompileGroup(b *testing.B) {
	for n := 0; n < b.N; n++ {
		regexp.Compile("^/f2(?:/(.*))/?$")
	}
}

func BenchmarkCompile(b *testing.B) {
	for n := 0; n < b.N; n++ {
		regexp.Compile("^/f2/.*/?$")
	}
}

func BenchmarkExecuteGroup(b *testing.B) {
	r, _ := regexp.Compile("^/f2(?:/(.*))/?$")
	for n := 0; n < b.N; n++ {
		r.MatchString("/f2/test/")
		r.MatchString("/different")
	}
}

func BenchmarkExecute(b *testing.B) {
	r, _ := regexp.Compile("^/f2/.*/?$")
	for n := 0; n < b.N; n++ {
		r.MatchString("/f2/test/")
		r.MatchString("/different")
	}
}
```

It's naïve, so i'm not sure how much we can trust it, but it shows that the group has a slight negative impact on performance (while having the same semantics, I think):

```
=== RUN   BenchmarkCompileGroup
BenchmarkCompileGroup
BenchmarkCompileGroup-10          461311              2570 ns/op         5000 B/op          61 allocs/op
=== RUN   BenchmarkCompile
BenchmarkCompile
BenchmarkCompile-10               542733              2152 ns/op         4240 B/op          53 allocs/op
=== RUN   BenchmarkExecuteGroup
BenchmarkExecuteGroup
BenchmarkExecuteGroup-10         7894125               152.6 ns/op          0 B/op           0 allocs/op
=== RUN   BenchmarkExecute
BenchmarkExecute
BenchmarkExecute-10              8257365               145.7 ns/op          0 B/op           0 allocs/op
PASS
```

This is a ~17% hit in compile time and ~5% in matching time.
Is there an easy way for us to prevent this change from adding capture groups to the output?

Skn0tt (approved):
Looks good. I left a point on performance, which can go into a follow-up though.

node/manifest.test.ts
eduardoboucas: We could remove capturing groups ourselves, since we're parsing the regular expression into an AST and transforming it. But I don't think that's a huge priority.

Skn0tt: Okay 👍 Maybe we can capture this in an issue, so we don't forget about it.

node/utils/urlpattern.ts
eduardoboucas: > At the very least we should pin an exact version

Totally. I actually meant to do this but ended up forgetting about it. Done in https://github.com/netlify/edge-bundler/pull/392/commits/b308434058e725cef3be1a63377fed508a46237e.

> My concern here is that we're opening ourselves to the possibility of the internals of the polyfill changing, as it's not a public API

If the internals change and the property we're using disappears or gets renamed, our own tests would catch it.

I'm not opposed to using our own fork if this makes people more comfortable — in fact, we do already have [a fork](https://github.com/netlify/urlpattern-polyfill), which I created in order to [push a change upstream](https://github.com/kenchris/urlpattern-polyfill/pull/111).

But if we go down that path, I would rather keep it on its own repo from where we can easily merge upstream changes, run the tests, etc. Not sure what we get by vendoring it in with Edge Bundler?

eduardoboucas:
> Put the change behind a flag, just to be safe. I'll deploy + start rolling this out tomorrow.

LGTM!

Skn0tt (approved):

PR #394, opened 2023-05-11
fix: remove del package

danez:
The package can be replaced by `fs.rm`

ref: https://github.com/netlify/cli/issues/5699


lukasholzer (approved):
LGTM!

PR #402, opened 2023-05-22
Add `excludedPath` support to ISC & TOML

Skn0tt:
This PR works towards https://github.com/netlify/pod-dev-foundations/issues/471. It implements the design I proposed in https://github.com/netlify/pod-dev-foundations/issues/471#issuecomment-1557109820.

### DIFF
--- a/node/bundler.test.ts
+++ b/node/bundler.test.ts
@@ -342,14 +342,15 @@ test('Loads declarations and import maps from the deploy configuration and in-so
 
   const manifestFile = await readFile(resolve(distPath, 'manifest.json'), 'utf8')
   const manifest = JSON.parse(manifestFile)
-  const { bundles, function_config: functionConfig } = manifest
+  const { bundles, function_config: functionConfig, routes } = manifest
   expect(bundles.length).toBe(1)
   expect(bundles[0].format).toBe('eszip2')
   expect(generatedFiles.includes(bundles[0].asset)).toBe(true)
 
   // respects excludedPath from deploy config
+  expect(routes[0].excluded_patterns).toEqual(['^/func2/skip/?$'])
+
   expect(functionConfig.func2).toEqual({
-    excluded_patterns: ['^/func2/skip/?$'],
     name: 'Function two',
     generator: '@netlify/fake-plugin@1.0.0',
   })
--- a/node/declaration.ts
+++ b/node/declaration.ts
@@ -13,7 +13,7 @@ interface BaseDeclaration {
 
 type DeclarationWithPath = BaseDeclaration & {
   path: Path
-  excludedPath?: Path
+  excludedPath?: Path | Path[]
 }
 
 type DeclarationWithPattern = BaseDeclaration & {
--- a/node/manifest.test.ts
+++ b/node/manifest.test.ts
@@ -167,6 +167,52 @@ test('Filters out internal in-source configurations in user created functions', 
   })
 })
 
+test('excludedPath from ISC goes into function_config, TOML goes into routes', () => {
+  const functions = [{ name: 'customisation', path: '/path/to/customisation.ts' }]
+  const declarations: Declaration[] = [
+    { function: 'customisation', path: '/showcases/*' },
+    { function: 'customisation', path: '/checkout/*', excludedPath: ['*/terms-and-conditions'] },
+  ]
+  const userFunctionConfig: Record<string, FunctionConfig> = {
+    customisation: {
+      excludedPath: ['*.css', '*.jpg'],
+    },
+  }
+  const internalFunctionConfig: Record<string, FunctionConfig> = {}
+  const manifest = generateManifest({
+    bundles: [],
+    declarations,
+    functions,
+    userFunctionConfig,
+    internalFunctionConfig,
+  })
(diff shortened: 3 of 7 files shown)


jenae-janzen:
👋 Trying to understand the context around this issue... wow 😅 What's the difference between `excludedPath` and `excludedPattern`? I see both referenced intermittently

node/manifest.test.ts
```diff
@@ -171,6 +167,52 @@ test('Filters out internal in-source configurations in user created functions',
...
+test('excludedPath from ISC goes into function_config, TOML goes into routes', () => {
+  const functions = [{ name: 'customisation', path: '/path/to/customisation.ts' }]
+  const declarations: Declaration[] = [
+    { function: 'customisation', path: '/showcases/*' },
+    { function: 'customisation', path: '/checkout/*', excludedPath: ['/*/terms-and-conditions'] },
+  ]
+  const userFunctionConfig: Record<string, FunctionConfig> = {
+    customisation: {
+      excludedPath: ['/*.css', '/*.jpg'],
+    },
```
jenae-janzen, line 179: If the user configures `excludedPath`, does it get saved as an `excluded_pattern`? According to this test, it does: both the declaration and the function config have an `excludedPath` that get converted to `excluded_pattern`. Is it correct to assume that whether the user sets `excludedPath` or `excludedPattern`, it will ultimately be treated the same way?

test/util.ts
```diff
@@ -52,8 +52,12 @@ const getRouteMatcher = (manifest: Manifest) => (candidate: string) =>
...
     }
 
-    const excludedPattern = manifest.function_config[route.function].excluded_patterns
-    const isExcluded = excludedPattern.some((pattern) => new RegExp(pattern).test(candidate))
+    if (route.excluded_patterns.some((pattern) => new RegExp(pattern).test(candidate))) {
+      return false
+    }
+
+    const excludedPatterns = manifest.function_config[route.function]?.excluded_patterns ?? []
+    const isExcluded = excludedPatterns.some((pattern) => new RegExp(pattern).test(candidate))
```
lukasholzer, line 60: What if the regex is invalid then it will throw here with an `Uncaught SyntaxError`.

Should we handle this case?

node/manifest.test.ts
lukasholzer: I've understood it the same way.

lukasholzer (commented):
Can you maybe add some tests for invalid regular expressions or patterns that don't start with a `^` and don't end with a `$`?

Skn0tt:
> 👋 Trying to understand the context around this issue... wow 😅

Yeah, I think we got side-tracked a bit in the discussion in the issue 😅

> What's the difference between excludedPath and excludedPattern? I see both referenced intermittently

They both achieve the same thing, but `excludedPath` is interpreted as a glob pattern, while `excludedPattern` is interpreted as a regex. This is only a user-facing distinction, though - under the hood, we compile the glob pattern into a regex. That's why you only see `excluded_pattern` referenced in the manifest.

node/manifest.test.ts
Skn0tt: Yes, that's correct. Context in https://github.com/netlify/edge-bundler/pull/402#issuecomment-1560570162.

node/bundler.test.ts
```diff
@@ -342,14 +342,15 @@ test('Loads declarations and import maps from the deploy configuration and in-so
...
+  const { bundles, function_config: functionConfig, routes } = manifest
   expect(bundles.length).toBe(1)
   expect(bundles[0].format).toBe('eszip2')
   expect(generatedFiles.includes(bundles[0].asset)).toBe(true)
 
   // respects excludedPath from deploy config
+  expect(routes[0].excluded_patterns).toEqual(['^/func2/skip/?$'])
+
   expect(functionConfig.func2).toEqual({
-    excluded_patterns: ['^/func2/skip/?$'],
```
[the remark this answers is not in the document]
Skn0tt: Right! Edge Functions can serve different routes, and so we make the distinction between "route config" and "function config". Route Config only applies to executions of a function that match that route, while function config will be applied to *all* executions, no matter the route.

node/declaration.ts
```diff
@@ -13,7 +13,7 @@ interface BaseDeclaration {
 
 type DeclarationWithPath = BaseDeclaration & {
   path: Path
-  excludedPath?: Path
+  excludedPath?: Path | Path[]
 }
 
 type DeclarationWithPattern = BaseDeclaration & {
```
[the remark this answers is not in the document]
Skn0tt: absolutely! done in 9e92fc81623dda07a0640fc02b06ab4c7cbed7cf

test/util.ts
Skn0tt: I'm unsure about that. This is unrelated to the PR, so let's open an issue and address it in another PR :)

Skn0tt:
> Can you maybe add some tests for invalid regular expressions or patterns that don't start with a ^ and don't end with a $?

Let's open an issue for that and address it in a different PR.

jenae-janzen (approved):
LGTM!

PR #417, opened 2023-06-08
chore: use vitest v8 coverage provider  and enable coverage

danez:
- vitest 0.32 changed the default coverage provider from c8 to v8, so let's use it.
- I also I noticed that coverage wasn't even enabled on CI.
- Also split up workflows to not run lint on every arch, it is enough to run it once on latest lts version of Node.js
- Noticed that the tests do not work, because the minimum version for deno is `1.30`, but the bootstrap is using features of deno `1.31+` https://github.com/netlify/edge-functions-bootstrap/commit/fa7e337bb903be575f5bab2c3501546d64c9c73f#diff-619f5f90a26baa69bacaaf654cbdf11eddc272ad34226a796112e3a1ec64d24bR2 //cc @Skn0tt 
- the tests though failed completely with deno `1.31`, so I set the minimum to `1.32` which now works
- test deno `1.32` (the minimum) as well as the latest version of deno

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -29,7 +29,7 @@ jobs:
         os: [ubuntu-latest]
         node-version: ['14.18.0', '*']
         # Must include the minimum deno version from the `DENO_VERSION_RANGE` constant in `node/bridge.ts`.
-        deno-version: ['v1.32.0', 'v1.x']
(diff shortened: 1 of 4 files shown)


eduardoboucas (approved):
Nice work!

eduardoboucas (approved):

PR #427, opened 2023-06-30
fix: parseConfig stumbling over `globalThis.Netlify` usage in global scope

Skn0tt:
When a function accesses `globalThis.Netlify`, the scanning for function config currently fails because it's not defining `globalThis.Netlify`.

This PR fixes that.

### DIFF
--- a/deno/config.ts
+++ b/deno/config.ts
@@ -1,6 +1,9 @@
 const [functionURL, collectorURL, rawExitCodes] = Deno.args
 const exitCodes = JSON.parse(rawExitCodes)
 
+// TODO: naïve, should this be imported from https://edge.netlify.com/bootstrap/globals.ts instead?
+globalThis.Netlify = { env: Deno.env }
+
 let func
 
 try {
--- a/deno/config.ts
+++ b/deno/config.ts
@@ -1,8 +1,8 @@
 const [functionURL, collectorURL, bootstrapURL, rawExitCodes] = Deno.args
 const exitCodes = JSON.parse(rawExitCodes)
 
-const { env } = await import(new URL('./globals.ts', bootstrapURL).toString())
-globalThis.Netlify = { env }
+const { Netlify } = await import(bootstrapURL)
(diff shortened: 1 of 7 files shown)


deno/config.ts
```diff
@@ -1,6 +1,9 @@
-const [functionURL, collectorURL, rawExitCodes] = Deno.args
+const [functionURL, collectorURL, bootstrapURL, rawExitCodes] = Deno.args
 const exitCodes = JSON.parse(rawExitCodes)
 
+const { env } = await import(new URL('./globals.ts', bootstrapURL).toString())
```
eduardoboucas, line 4: Doesn't this implicitly make the `globals.ts` an external API? Like, if you rename that file as part of a refactoring in the bootstrap repo and merge the PR, this would silently break?

Could we export `env` from `index-combined.ts`, which is a public entry point, and then import it from there?

Skn0tt: > Could we export env from index-combined.ts, which is a public entry point, and then import it from there?

That sounds like a good idea, yes. What do you think about exporting `Netlify` instead, so that we can do `globalThis.Netlify = Netlify`?

eduardoboucas: Sounds good!

Skn0tt: Addressed in https://github.com/netlify/edge-bundler/pull/427/commits/bf03f1992b301bb65d2e5862468824160854c109.

node/bundler.ts
```diff
@@ -49,6 +50,7 @@ const bundle = async (
     onBeforeDownload,
     systemLogger,
     internalSrcFolder,
+    bootstrapURL = 'https://edge.netlify.com/bootstrap/index-combined.ts',
```
Skn0tt, line 53: My gut feeling is that since we're only using this for the ISC check, it's not of the utmost importance that this comes from DevCycle. So I made it optional for now, and i'll open an issue for us to have buildbot pass this in. Once that's there, we can make it required here.

PR #440, opened 2023-07-25
feat!: correctly interpret `/**/*` paths

Skn0tt:
This came up over in this forums post: https://answers.netlify.com/t/please-clarify-documentation-for-edge-functions-excludedpath/97443

Given the path `/**/*.html`, i'd expect it to match both `/foo.html` and `/sub/foo.html`. With our current handling, only `/sub/foo.html` is matched however - i've demonstrated this in a test added in the first commit of this PR.

After reading into this, it turns out that the special meaning of `**` [was introduced "recently" in 2010](https://www.linuxjournal.com/content/globstar-new-bash-globbing-option) and is called `globstar`. We don't have it enabled for edge functions currently, which means that `/**/*.html` is equivalent to `/*/*.html`. This probably comes at a surprise to our customers.

This PR enables the globstar option. This is a breaking change to what we implemented, and i'm unsure what's the best way of rolling this out. Here's a matrix for how it'd change behaviour:

|  | matched paths, currently  | with `globstar` |
|- |-| -|
| `/*.html` | `/foo.html`, `/sub/foo.html` | `/foo.html` |
| `/**/*.html` | `/sub/foo.html` | `/foo.html`, `/sub/foo.html`|

This would essentially break sites that have an edge function at `/*`, so i'm really unsure if this is a change we can make now 🤔  Let's have a discussion on that.

### DIFF
--- a/node/config.test.ts
+++ b/node/config.test.ts
@@ -198,7 +198,7 @@ test('Loads function paths from the in-source `config` function', async () => {
   expect(routes[2]).toEqual({ function: 'framework-func1', pattern: '^/framework-func1/?$', excluded_patterns: [] })
   expect(routes[3]).toEqual({ function: 'user-func1', pattern: '^/user-func1/?$', excluded_patterns: [] })
   expect(routes[4]).toEqual({ function: 'user-func3', pattern: '^/user-func3/?$', excluded_patterns: [] })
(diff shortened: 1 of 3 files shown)


Skn0tt:
closing in favor of https://github.com/netlify/edge-bundler/pull/392

PR #453, opened 2023-08-16
feat: simplify `ImportMap`

eduardoboucas:

The current implementation of the `ImportMap` class is too complex to understand and too hard to extend.

I've extended the tests and I'll also add inline comments to add some colour to some of the changes.

This is an internal change, so no customer-facing behaviour should change.

### DIFF
--- a/node/import_map.ts
+++ b/node/import_map.ts
@@ -121,18 +121,24 @@ export class ImportMap {
     return filteredImports
   }
 
+  // Takes a `scopes` object and runs all imports through `filterImports`,
+  // omitting any scopes for which there are no imports.
   filterScopes(scopes?: ParsedImportMap['scopes']) {
-    if (scopes === undefined) {
-      return {}
+    const filteredScopes: Record<string, Imports> = {}
+
+    if (scopes !== undefined) {
+      Object.keys(scopes).forEach((url) => {
+        const imports = this.filterImports(scopes[url])
+
+        if (Object.keys(imports).length === 0) {
+          return
+        }
+
+        filteredScopes[url] = imports
+      })
     }
 
-    const resolvedScopes: Record<string, Imports> = {}
-
-    Object.keys(scopes || {}).forEach((url) => {
-      resolvedScopes[url] = this.filterImports(scopes[url])
-    })
-
-    return resolvedScopes
+    return filteredScopes
   }
 
   // Returns the import map as a plain object, with any relative paths resolved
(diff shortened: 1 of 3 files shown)


node/import_map.ts
```diff
@@ -173,31 +233,3 @@ class ImportMap {
     await fs.writeFile(path, JSON.stringify(contents))
   }
 }
-
-const readFile = async (path: string, logger: Logger): Promise<ImportMapSource> => {
```
eduardoboucas, line 177: This is now a static method in the `ImportMap` class.

node/import_map.ts
```diff
@@ -64,79 +104,53 @@ class ImportMap {
...
-      await this.addFile(path, logger)
-    }
-  }
-
-  getContents(basePath?: string, prefix?: string) {
+  // Returns the import map as a plain object, with any relative paths resolved
+  // to full URLs. It takes an optional `prefixes` object that specifies a list
+  // of prefixes to replace path prefixes (see `applyPrefixesToPath`). Prefixes
+  // will be applied on both `imports` and `scopes`.
+  getContents(prefixes: Record<string, string> = {}) {
```
eduardoboucas, line 148: Previously, `getContents` accepted a base path parameter, which doesn't really make sense because all paths are already absolute before this method returns.

What we want is the ability to transform the paths to use different prefixes (such as the `file:///root/` virtual root for ESZIPs), so this method now accepts an optional `prefixes` object. So far we have one prefix, but we'll likely have more in the future.

node/import_map.ts
```diff
@@ -14,47 +14,87 @@ const INTERNAL_IMPORTS = {
...
-    const { imports = {}, scopes = {} } = parsedImportMap
-    const resolvedImports = ImportMap.resolveImports(imports, basePath, prefix)
-    const resolvedScopes: Record<string, Imports> = {}
+  add(source: ImportMapFile) {
+    this.sources.push(source)
+  }
 
-    Object.keys(scopes).forEach((path) => {
-      const resolvedPath = ImportMap.resolvePath(new URL(path), basePath, prefix)
+  async addFile(path: string, logger: Logger) {
```
eduardoboucas, line 46: No change in this method, other than using `readFile` from the class.

node/import_map.ts
```diff
@@ -14,47 +14,87 @@ const INTERNAL_IMPORTS = {
...
-      resolvedScopes[resolvedPath] = ImportMap.resolveImports(scopes[path], basePath, prefix)
-    })
+    if (Object.keys(source.imports).length === 0) {
+      return
+    }
+
+    return this.add(source)
+  }
+
+  async addFiles(paths: (string | undefined)[], logger: Logger) {
```
eduardoboucas, line 56: No change in this method.

lukasholzer (approved):
LGTM!

PR #408, opened 2023-06-01
feat: add experimental support for npm modules

eduardoboucas:
⏳

PR #454, opened 2023-08-16
feat: add support for npm modules

eduardoboucas:

Adds support for importing npm modules as bare specifiers, with or without the `npm:` prefix. It uses esbuild to bundle all npm modules into JavaScript files that the ESZIP bundler can process.

Take the edge function below as an example.

```ts
import m1 from "module-1"
import m2 from "npm:module-2"
import m3 from "https://example.netlify/mod.ts"

export default async (req: Request) => {
  const body = m1(m2(m3("hello")))

  return new Response(body)
}
```

To bundle this function, the following will happen:

1. Traverse the code with esbuild to collect the names of the specifiers that should map to npm modules. We'll find `module-1` and `npm:module-2`.
2. Create the following "stub" files:

    ```js
    // stub-0.js
    export { default } from "module-1"; export * from "module-1";

    // stub-1.js
    export { default } from "module-2"; export * from "module-2";
    ```
3. Bundle the stub files with esbuild, leading to new `stub-0.js` and `stub-1.js` dist files containing a bundle with `module-1` and `module-2` along with all of their dependencies.
4. Add `module-1` and `npm:module-2` to the import map such that they map to the stub files when creating the ESZIP bundle

This is a better approach than https://github.com/netlify/edge-bundler/pull/408 (which this PR supersedes), because we keep the user code untouched. We simply bundle the dependencies that ESZIP can't natively handle.

The functionality is behind a feature flag so it should be fairly safe to roll out.


eduardoboucas:
Of course that Windows tests are failing... ⏳

node/npm_dependencies.ts
```diff
@@ -0,0 +1,147 @@
...
+  // stubs (such that a common module isn't bundled twice).
+  await build({
+    allowOverwrite: true,
+    bundle: true,
+    entryPoints,
+    format: 'esm',
+    nodePaths,
+    outdir: temporaryDirectory.path,
+    platform: 'node',
+    splitting: true,
```
Skn0tt, line 111: esbuild docs mention some known problems with this setting, how do we know if this works for us? https://esbuild.github.io/api/#splitting

node/npm_dependencies.ts
```diff
@@ -0,0 +1,147 @@
...
+
+      return { filePath, specifier }
+    }),
+  )
+  const entryPoints = ops.map(({ filePath }) => filePath)
+
+  // Bundle each of the stub files we've created. We'll end up with a compiled
+  // version of each of the stub files, plus any chunks of shared code between
+  // stubs (such that a common module isn't bundled twice).
+  await build({
```
Skn0tt, line 103: I believe in order for things to work with Deno, we need to prefix all Node.js stdlib imports with `node:`, so Deno knows that it should use the Node compat library. Does ESBuild do that for us? I don't think we have a unit test for it, but I think we should.

Skn0tt (commented):
The approach looks sound. Left a comment on node stdlib imports.

This seems like the sort of thing that will work fine for 90% of usecases, but is hard to debug in cases where it's not. Do we know why deno hasn't shipped their `npm:` Prefix to us? I'd much prefer that, over having to build something ourselves 😅

node/npm_dependencies.ts
eduardoboucas: Great call out. Sadly, esbuild can't do this for us automatically because it doesn't rewrite imports, but we can do that ourselves using an import map. See https://github.com/netlify/edge-bundler/pull/454/commits/99b95373e277c5f7428644fea4b07bec4c8b570b.

node/npm_dependencies.ts
eduardoboucas: Yeah, I was unsure about this myself. We don't have to use it. The other option is that we use a single entry point with all dependencies. The only disadvantage is that if a dependency is referenced in the code but not actually used, it won't get tree-shaked. But that's not a huge deal, I don't think.

I'd say we can use code splitting for an initial implementation and see if it causes problems. No strong opinions though.

Skn0tt (commented):
I see how this adds support for resolving node stdlib from within the transpiled packages, but I think it also adds that support to user code. I'm unsure if we want that. Is there a way to have that import map only apply to the transpiled files?

Skn0tt (approved):

minivan (approved):

PR #455, opened 2023-08-16
feat: add `path` to manifest

eduardoboucas:
Adds a `path` property to the manifest file containing the original path defined for the function, before any transformations to regular expressions.

Part of https://github.com/netlify/pod-dev-foundations/issues/459.

### DIFF
--- a/node/manifest.ts
+++ b/node/manifest.ts
@@ -18,6 +18,7 @@ interface Route {
   function: string
   pattern: string
   excluded_patterns: string[]
+  raw_pattern?: string
 }
 
 interface EdgeFunctionConfig {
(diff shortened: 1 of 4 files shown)


node/manifest.ts
```diff
@@ -141,6 +142,10 @@ const generateManifest = ({
       excluded_patterns: excludedPattern.map(serializePattern),
     }
 
+    if ('path' in declaration) {
+      route.raw_pattern = declaration.path
```
eduardoboucas, line 146: Not sure about the name. `path` or `pattern` would make sense here, but they're both being used already. `raw_pattern` is the best I could come up with, but I'm happy to take suggestions.

Worth noting that this is an internal field only, so not massively important.

Skn0tt (approved):

PR #458, opened 2023-08-22
feat: match on http methods

Skn0tt:
Part of https://github.com/netlify/pod-dev-foundations/issues/579.

### DIFF
--- a/node/manifest.ts
+++ b/node/manifest.ts
@@ -92,7 +92,8 @@ const addExcludedPatterns = (
 }
 
 /**
- * Turns 'get' into ['GET']
+ * Normalizes method names into arrays of uppercase strings.
+ * (e.g. "get" becomes ["GET"])
  */
 const normalizeMethods = (method: undefined | string | string[]) => {
   if (!method) return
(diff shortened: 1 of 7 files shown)


node/manifest.ts
```diff
@@ -90,6 +91,16 @@ const addExcludedPatterns = (
   }
 }
 
+/**
+ * Turns 'get' into ['GET']
```
eduardoboucas, line 95: mega nit

```suggestion
 * Normalizes method names into arrays of uppercase strings.
 * (e.g. "get" becomes ["GET"])
```

node/manifest.ts
```diff
@@ -90,6 +91,16 @@ const addExcludedPatterns = (
...
+ */
+const normalizeMethods = (method: undefined | string | string[]) => {
+  if (!method) return
+
+  const methods = Array.isArray(method) ? method : [method]
+  return methods.map((method) => method.toUpperCase())
```
eduardoboucas, line 101: We take this from user code that might not even be typed, so if someone uses a non-string value in the `method` property, we'll throw a bit of a cryptic error when trying to call `.toUpperCase()`?

Skn0tt: good catch! added a type check in d9c00080dfd053d388748948a5311ba14eceedbf.

eduardoboucas (approved):

PR #464, opened 2023-08-28
fix: hide stack trace on syntax errors

Skn0tt:
Closes https://github.com/netlify/pod-dev-foundations/issues/580.

### DIFF
--- a/deno/bundle.ts
+++ b/deno/bundle.ts
@@ -3,4 +3,14 @@ import { writeStage2 } from './lib/stage2.ts'
 const [payload] = Deno.args
 const { basePath, destPath, externals, functions, importMapData } = JSON.parse(payload)
 
-await writeStage2({ basePath, destPath, externals, functions, importMapData })
+try {
+  await writeStage2({ basePath, destPath, externals, functions, importMapData })
+} catch (error) {
+  if (error instanceof Error) {
+    if (error.message.includes("The module's source code could not be parsed")) {
+      delete error.stack
+    }
+  }
+
+  throw error
+}
--- a/deno/bundle.ts
+++ b/deno/bundle.ts
@@ -6,10 +6,8 @@ const { basePath, destPath, externals, functions, importMapData } = JSON.parse(p
 try {
   await writeStage2({ basePath, destPath, externals, functions, importMapData })
 } catch (error) {
-  if (error instanceof Error) {
-    if (error.message.includes("The module's source code could not be parsed")) {
-      delete error.stack
-    }
+  if (error instanceof Error && error.message.includes("The module's source code could not be parsed")) {
+    delete error.stack
   }
 
   throw error
--- a/node/bundle_error.ts
+++ b/node/bundle_error.ts
@@ -1,3 +1,5 @@
+import type { ExecaError } from 'execa'
+
 interface BundleErrorOptions {
   format: string
 }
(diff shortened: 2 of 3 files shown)


node/bundler.test.ts
```diff
@@ -102,6 +103,15 @@ test('Adds a custom error property to user errors during bundling', async () =>
...
   } catch (error) {
     expect(error).toBeInstanceOf(BundleError)
+    expect((error as BundleError).message).toMatchInlineSnapshot(`
+      "error: Uncaught (in promise) Error: The module's source code could not be parsed: Unexpected eof at file:///root/functions/func1.ts:1:27
+
+        export default async () => 
+                                  ~
+            const ret = new Error(getStringFromWasm0(arg0, arg1));
+                        ^
+          at <anonymous> (file:///Users/skn0tt/dev/netlify/edge-bundler/deno/vendor/deno.land/x/eszip@v0.40.0/eszip_wasm.generated.js:417:19)"
```
eduardoboucas, line 113: Your local paths will be a problem when matching the snapshot.

eduardoboucas (approved):
Nice one! You have a test failure, but the code looks good to me.

node/bundler.test.ts
```diff
@@ -86,8 +86,9 @@ test('Uses the vendored eszip module instead of fetching it from deno.land', asy
   await cleanup()
 })
 
-test('Adds a custom error property to user errors during bundling', async () => {
-  expect.assertions(2)
+test.only('Adds a custom error property to user errors during bundling', async () => {
```
eduardoboucas, line 89:
```suggestion
test('Adds a custom error property to user errors during bundling', async () => {
```

deno/bundle.ts
```diff
@@ -3,4 +3,14 @@ import { writeStage2 } from './lib/stage2.ts'
 const [payload] = Deno.args
 const { basePath, destPath, externals, functions, importMapData } = JSON.parse(payload)
 
-await writeStage2({ basePath, destPath, externals, functions, importMapData })
+try {
+  await writeStage2({ basePath, destPath, externals, functions, importMapData })
+} catch (error) {
+  if (error instanceof Error) {
+    if (error.message.includes("The module's source code could not be parsed")) {
```
eduardoboucas, line 10: Could we not have a single `if` here?

```js
if (error instanceof Error && error.message.includes("The module's source code could not be parsed")) {
  // (...)
}
```

Skn0tt: done in 10bbe187cd6b8b727ce0d9fedc2ecf2996350ad2

node/bundle_error.ts
```diff
@@ -30,6 +32,11 @@ class BundleError extends Error {
  */
 const wrapBundleError = (input: unknown, options?: BundleErrorOptions) => {
   if (input instanceof Error) {
+    if (input.message.includes("The module's source code could not be parsed")) {
```
eduardoboucas, line 35: Same here, if you don't mind.

Skn0tt: I don't think that works - we want to wrap it in a `BundleError`, even if the message is something different

eduardoboucas: Oops, sorry. Didn't look at this properly.

eduardoboucas (approved):

PR #471, opened 2023-09-07
fix: support import maps in npm module resolution

eduardoboucas:

This is a follow-up to #454 addressing a couple of issues I found when testing this in production.

The main change is that the esbuild plugin needs to be aware of user-defined import maps, otherwise it will fail to resolve specifiers that have been remapped.

I ended up making other changes, which I'll describe using inline comments.

### DIFF
--- a/node/import_map.ts
+++ b/node/import_map.ts
@@ -76,6 +76,26 @@ export class ImportMap {
     )
   }
 
+  static convertImportsToURLObjects(imports: Imports) {
+    return Object.entries(imports).reduce(
+      (acc, [key, value]) => ({
+        ...acc,
+        [key]: new URL(value),
+      }),
+      {} as Record<string, URL>,
+    )
+  }
+
+  static convertScopesToURLObjects(scopes: Record<string, Imports>) {
+    return Object.entries(scopes).reduce(
+      (acc, [key, value]) => ({
+        ...acc,
+        [key]: ImportMap.convertImportsToURLObjects(value),
+      }),
+      {} as Record<string, Record<string, URL>>,
+    )
+  }
+
   // Applies a list of prefixes to a given path, returning the replaced path.
   // For example, given a `path` of `file:///foo/bar/baz.js` and a `prefixes`
   // object with `{"file:///foo/": "file:///hello/"}`, this method will return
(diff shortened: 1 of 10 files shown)


node/import_map.ts
```diff
@@ -180,6 +200,18 @@ export class ImportMap {
     }
   }
 
+  // The same as `getContents`, but the URLs are represented as URL objects
+  // instead of strings. This is compatible with the `ParsedImportMap` type
+  // from the `@import-maps/resolve` library.
+  getContentsWithURLObjects(prefixes: Record<string, string> = {}) {
```
eduardoboucas, line 206: This is a bit annoying, but we're using `@import-maps/resolve` to resolve import maps, and it expects `URL` objects, whereas `getContents()` returns URLs as strings.

I suppose I could've changed `getContents()` to use `URL` objects, since they should be converted to strings automatically when marshalled to JSON. I felt that creating a separate method would be an easier first take, but I might still do that as a follow-up.

node/npm_dependencies.ts
```diff
@@ -62,34 +134,48 @@ export const vendorNPMSpecifiers = async (basePath: string, functions: string[],
   // resolution logic won't work.
   const nodePaths = [path.join(basePath, 'node_modules')]
 
+  // We need to create some files on disk, which we don't want to write to the
+  // project directory. If a custom directory has been specified, which happens
+  // only in tests, we use it. Otherwise, create a random temporary directory.
+  const temporaryDirectory = directory ? { path: directory } : await tmp.dir()
```
eduardoboucas, line 140: Moved this up so that we can pass it as `outdir` to the first `build` call. That shouldn't be needed, because we're setting `write: false`, so we won't write any files to disk anyway. But esbuild still complains, which I think is a bug.

node/npm_dependencies.ts
```diff
@@ -62,34 +134,48 @@ export const vendorNPMSpecifiers = async (basePath: string, functions: string[],
...
+      nodePaths,
+      outdir: temporaryDirectory.path,
+      platform: 'node',
+      plugins: [getDependencyTrackerPlugin(specifiers, importMap.getContentsWithURLObjects(), pathToFileURL(basePath))],
+      write: false,
+    })
+  } catch (error) {
+    logger.system('Could not track dependencies in edge function:', error)
+    logger.user(
+      'An error occurred when trying to scan your Edge Functions for npm modules, which is an experimental feature. If you are loading npm modules, please share the errors above in https://ntl.fyi/edge-functions-npm. If you are not loading npm modules, you can ignore this message.',
```
eduardoboucas, line 160: This will point to a forums post where we explain the feature and the fact that it's experimental.

Skn0tt (approved):
Looks good! Thanks for all the comments, super helpful in reviewing.

Skn0tt (approved):

PR #472, opened 2023-09-08
feat: remove support for `npm:` prefix

eduardoboucas:

This is a follow-up to #454. With the goal of cutting scope for the initial release, this PR removes support for the `npm:` prefix, letting users load npm modules only when using an un-prefixed bare specifier.

This is because using the `npm:` prefix involves supporting the module version as part of the specifier, which requires additional work. We might still add this, but we may decide to exclude it from the initial release.

Part of https://github.com/netlify/pod-dev-foundations/issues/578.

### DIFF
--- a/node/npm_dependencies.ts
+++ b/node/npm_dependencies.ts
@@ -18,9 +18,9 @@ const require = createRequire(import.meta.url)
 // Workaround for https://github.com/evanw/esbuild/issues/1921.
 const banner = {
   js: `
-  import {createRequire as ___nfyCreateRequire} from "module";
-  import {fileURLToPath as ___nfyFileURLToPath} from "url";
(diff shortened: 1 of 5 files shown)


node/npm_dependencies.ts
```diff
@@ -170,13 +165,13 @@ export const vendorNPMSpecifiers = async ({
     'You are using npm modules in Edge Functions, which is an experimental feature. Learn more at https://ntl.fyi/edge-functions-npm.',
   )
 
-  // To bundle an entire module and all its dependencies, we create a stub file
+  // To bundle an entire module and all its dependencies, create a barrel file
```
jenae-janzen, line 168: TIL what a barrel file is!

jenae-janzen (approved):

PR #475, opened 2023-09-11
feat!: support npm modules when serving

eduardoboucas:

Another follow-up to #454, supporting npm modules in the `serve` entry point.

This is technically a breaking change since `serve` now expects new `basePath` and `servePath` parameters.

### DIFF
--- a/node/server/server.ts
+++ b/node/server/server.ts
@@ -1,21 +1,22 @@
-import { tmpName } from 'tmp-promise'
-
 import { DenoBridge, OnAfterDownloadHook, OnBeforeDownloadHook, ProcessRef } from '../bridge.js'
 import { getFunctionConfig, FunctionConfig } from '../config.js'
 import type { EdgeFunction } from '../edge_function.js'
 import { generateStage2 } from '../formats/javascript.js'
 import { ImportMap } from '../import_map.js'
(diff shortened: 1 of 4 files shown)


node/server/server.ts
```diff
@@ -73,12 +89,13 @@ const prepareServer = ({
       // no-op
     }
 
-    const bootstrapFlags = ['--port', port.toString()]
+    const extraDenoFlags = [`--import-map=${importMap.toDataURL()}`]
```
eduardoboucas, line 92: The main change here is that rather than computing the import map once the server starts, we do it every time there's a file change. This is because the import map may change when the code adds or remove npm modules.

I've added the `clone()` method to `ImportMap` so that we can keep an import map around with the base entries, and then clone and extend it with new entries on file changes.

node/server/server.ts
```diff
@@ -141,6 +165,7 @@ const serve = async ({
   onAfterDownload,
   onBeforeDownload,
   port,
+  servePath,
```
eduardoboucas, line 168: The new `servePath` will be used to write any temporary files, such as the stage 2 file, rather than using a temporary directory.

While working on this, I noticed that we weren't previously cleaning up the temporary directories we created. We address that here.

jenae-janzen (approved):

PR #481, opened 2023-09-25
feat: return `features` from server

eduardoboucas:

This is a follow-up to #475, where we started logging a warning message whenever we detect that npm modules are being used. When used in the CLI, this means that this message will be shown every time an edge function is saved and a new server is started, which is not ideal.

In this PR, I'm removing that warning from Edge Bundler and instead making the server return a generic `features` object that informs consumers about specific features that are being used. When npm modules are detected, this object will be `{npmModules: true}`.

With this, the CLI can find this information and be in charge of logging the warning, with the added benefit that it can decide to only log it once. It also lets the CLI log the message with whatever formatting makes more sense in the context of Netlify Dev.

It also wraps the npm vendoring in a feature flag.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -29,7 +29,7 @@ jobs:
         os: [ubuntu-latest]
         node-version: ['14.18.0', '*']
         # Must include the minimum deno version from the `DENO_VERSION_RANGE` constant in `node/bridge.ts`.
-        deno-version: ['v1.32.5', 'v1.x']
+        deno-version: ['v1.37.0', 'v1.x']
         include:
           - os: macOS-latest
             node-version: '*'
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -29,7 +29,7 @@ jobs:
         os: [ubuntu-latest]
         node-version: ['14.18.0', '*']
         # Must include the minimum deno version from the `DENO_VERSION_RANGE` constant in `node/bridge.ts`.
-        deno-version: ['v1.37.0', 'v1.x']
+        deno-version: ['v1.32.5', 'v1.x']
         include:
           - os: macOS-latest
             node-version: '*'
--- a/node/npm_dependencies.ts
+++ b/node/npm_dependencies.ts
@@ -161,10 +161,6 @@ export const vendorNPMSpecifiers = async ({
     return
   }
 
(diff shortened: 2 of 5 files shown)


.github/workflows/workflow.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         os: [ubuntu-latest]
         node-version: ['14.18.0', '*']
         # Must include the minimum deno version from the `DENO_VERSION_RANGE` constant in `node/bridge.ts`.
-        deno-version: ['v1.32.5', 'v1.x']
+        deno-version: ['v1.37.0', 'v1.x']
```
eduardoboucas, line 32: Updating to the latest bootstrap means that we now rely on `Deno.serve` to run tests, which wasn't available on 1.32.5.

node/npm_dependencies.ts
```diff
@@ -69,7 +69,7 @@ export const getDependencyTrackerPlugin = (
         return { external: true }
       }
 
-      const isLocalImport = specifier.startsWith(path.sep) || specifier.startsWith('.')
+      const isLocalImport = specifier.startsWith(path.sep) || specifier.startsWith('.') || path.isAbsolute(specifier)
```
eduardoboucas, line 72: Fixes an issue where importing an absolute directory on Windows was not being treated as a local import.

minivan (approved):

node/server/server.ts
```diff
@@ -64,17 +67,22 @@ const prepareServer = ({
...
-      basePath,
-      directory: distDirectory,
-      functions: functions.map(({ path }) => path),
-      importMap,
-      logger,
-    })
 
-    if (vendor) {
-      importMap.add(vendor.importMap)
+    if (featureFlags?.edge_functions_npm_modules) {
```
Skn0tt, line 73: I'm assuming that you're planning on statically setting this flag from CLI, or do you want to roll this out via DevCycle?

Skn0tt (approved):
Looks good 👍 Do you think it makes sense to return the same `features` object from `bundle`?

node/server/server.ts
eduardoboucas: Using DevCycle.

eduardoboucas:
> Do you think it makes sense to return the same `features` object from `bundle`?

Yes! I'll do that in a follow-up.

PR #483, opened 2023-09-28
fix: detect .mjs files

Skn0tt:
We're supporting `.mjs` files in Netlify Functions, so we should do the same for Edge Functions.

I'm not adding `.cjs` here, since Edge Functions need to be ES Modules.


eduardoboucas (commented):
Any idea why tests are failing?

minivan (approved):

PR #493, opened 2023-10-09
feat: allow injecting user-facing logger

Skn0tt:
We should handle logging within the entrypoints of our systems, not within the satellite packages. This PR allows injecting a logging provider for user-facing logs.

### DIFF
--- a/node/bridge.ts
+++ b/node/bridge.ts
@@ -56,7 +56,7 @@ class DenoBridge {
     this.cacheDirectory = options.cacheDirectory ?? getPathInHome('deno-cli')
     this.debug = options.debug ?? false
     this.denoDir = options.denoDir
-    this.logger = options.logger ?? getLogger(undefined, options.debug)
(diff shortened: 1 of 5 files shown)


eduardoboucas (approved):
Nice!

node/bridge.ts
```diff
@@ -56,7 +56,7 @@ class DenoBridge {
     this.cacheDirectory = options.cacheDirectory ?? getPathInHome('deno-cli')
     this.debug = options.debug ?? false
     this.denoDir = options.denoDir
-    this.logger = options.logger ?? getLogger(undefined, options.debug)
+    this.logger = options.logger ?? getLogger(undefined, undefined, options.debug)
```
eduardoboucas, line 59: nit: Might be nice to make `getLogger` receive an object so that we don't have to supply a bunch of `undefined`.

Totally not a blocker, though.

eduardoboucas (approved):

PR #494, opened 2023-10-09
fix: NPM bundling should use ESM format

Skn0tt:
When we try detecting imported NPM modules, we use ESBuild. Right now, it sometimes fails that tracing process because of code-level errors. One specific example for that is top-level await, which is only available in ES Modules. This PR fixes this specific instance by switching the bundling format for NPM modules tracing from the (default) CJS format to ESM.

### DIFF
--- a/node/npm_dependencies.ts
+++ b/node/npm_dependencies.ts
@@ -148,6 +148,7 @@ export const vendorNPMSpecifiers = async ({
       platform: 'node',
       plugins: [getDependencyTrackerPlugin(specifiers, importMap.getContentsWithURLObjects(), pathToFileURL(basePath))],
       write: false,
+      format: 'esm',
     })
   } catch (error) {
     logger.system('Could not track dependencies in edge function:', error)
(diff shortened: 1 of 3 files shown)


eduardoboucas (approved):

node/npm_dependencies.ts
```diff
@@ -148,6 +148,7 @@ export const vendorNPMSpecifiers = async ({
       platform: 'node',
       plugins: [getDependencyTrackerPlugin(specifiers, importMap.getContentsWithURLObjects(), pathToFileURL(basePath))],
       write: false,
+      format: 'esm',
```
Skn0tt, line 151: The more general problem is that ESBuild performs all kinds of syntax checks on the code that it parses. For example, it will stumble over this snippet:

```ts
function deadCodeThatIsNeverCalled() {
  const foo = ""
  foo = "hello"
}
```

In my REPL, this didn't throw until calling it - so if it's really dead code, then it's OK to include it in the bundle. ESBuild will throw a warning, though.

This might be a problem, because it prevents code with false-positives like this from using NPM modules 🤔

Skn0tt: (it's not a made-up scenario, i've seen instances of this error in Humio. stuff for a follow-up though)

PR #497, opened 2023-10-10
fix: mute esbuild while parsing for NPM modules

Skn0tt:
Parsing NPM modules might fail. We already have a user-facing log line for that, so we don't need ESBuild's output on top of it. This PR mutes esbuild, and directs the generated warnings and errors to system logs.


Skn0tt:
Wanted to get this out today. I can still write a test for this tomorrow, if we think that's necessary

eduardoboucas (dismissed):

Skn0tt:
Verified locally that this does mute the ESBuild errors. Updated the error message to not refer to the esbuild output.

minivan (approved):

PR #499, opened 2023-10-13
feat: trace npm modules with NFT

eduardoboucas:

Trace npm modules with NFT instead of esbuild.

https://netlify.slack.com/archives/C053C52PZKR/p1697095049145689?thread_ts=1696941439.424519&cid=C053C52PZKR


node/npm_dependencies.ts
```diff
@@ -14,6 +16,7 @@ import { Logger } from './logger.js'
 
 const builtinModulesSet = new Set(builtinModules)
 const require = createRequire(import.meta.url)
+const TYPESCRIPT_EXTENSIONS = new Set(['.ts', '.cts', 'mjs'])
```
Skn0tt, line 19: Is `mjs` in here on purpose? Should it be `.mts`?

node/npm_dependencies.ts
```diff
@@ -138,45 +230,30 @@ export const vendorNPMSpecifiers = async ({
...
+  const { modulesWithExtraneousFiles, npmSpecifiers } = await getNPMSpecifiers(
+    basePath,
+    functions,
+    importMap.getContentsWithURLObjects(),
+  )
+
+  if (modulesWithExtraneousFiles.length !== 0) {
     logger.user(
-      'An error occurred when trying to scan your edge functions for npm modules, which is an experimental feature. If you are loading npm modules, please share the link to this deploy in https://ntl.fyi/edge-functions-npm. If you are not loading npm modules, you can ignore this message.',
+      `These npm modules, imported directly or indirectly by an edge function, appear to dynamically import files at runtime, which is currently not supported (https://ntl.fyi/edge-npm): ${modulesWithExtraneousFiles.join(
```
Skn0tt, line 241: Nice that we can detect this now!

Skn0tt (commented):
Looking good! There's one copy&paste error in there, I believe.

eduardoboucas:
@Skn0tt Okay, tests are finally working! I've also addressed your comments from the review. Also removed the dependency tracing esbuild plugin, which we're no longer using.

node/npm_dependencies.ts
```diff
@@ -1,19 +1,18 @@
...
 import tmp from 'tmp-promise'
 
-import { nodePrefix, npmPrefix } from '../shared/consts.js'
-
 import { ImportMap } from './import_map.js'
 import { Logger } from './logger.js'
 
-const builtinModulesSet = new Set(builtinModules)
-const require = createRequire(import.meta.url)
+const TYPESCRIPT_EXTENSIONS = new Set(['.ts', '.cts', 'mts'])
```
Skn0tt, line 15:
```suggestion
const TYPESCRIPT_EXTENSIONS = new Set(['.ts', '.cts', '.mts'])
```

eduardoboucas: Jesus christ, this one is haunted. Thanks!

Skn0tt (dismissed):

lukasholzer (approved):
LGTM

eduardoboucas:
I'm really sorry for another bump, but I pushed a final change in https://github.com/netlify/edge-bundler/pull/499/commits/f56218eb43bdda43e6536ac546c92dc5048f77a3. Rather than making Edge Bundler do the logging, I think returning the specifiers back to the CLI would allow us to handle the messaging a lot better.

For example, we could start displaying this notice when we load/reload an edge function.

@Skn0tt @lukasholzer could I get another ✅, please?

Skn0tt (approved):

PR #505, opened 2023-10-19
feat: detect Typescript typings for NPM modules and reference them from barrel files

Skn0tt:
Closes https://linear.app/netlify/issue/COM-53/load-types-when-using-npm-modules-in-edge-functions.

When generating barrel files for NPM modules, type information can get lost. To prevent that, we're adding triple-slash typescript directive, that editors can use to resolve typing information. More info on that [in the TypeScript docs](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-types-).

I couldn't find a library to do this for us online, sadly :/ So I had to implement most of the logic myself. It's a bit convoluted, i'm very open to improvements here.


node/npm_dependencies.ts
```diff
@@ -72,14 +106,14 @@ const getNPMSpecifiers = async (basePath: string, functions: string[], importMap
       return nftResolve(specifier, ...args)
     },
   })
-  const npmSpecifiers = new Set<string>()
+  const npmSpecifiers: Record<string, { types?: string }> = {}
   const npmSpecifiersWithExtraneousFiles = new Set<string>()
 
-  reasons.forEach((reason, filePath) => {
+  for (const [filePath, reason] of reasons.entries()) {
```
Skn0tt, line 112: making a for-loop out of this, so we can use `await` down below

minivan (approved):

node/npm_dependencies.ts
```diff
@@ -14,6 +15,39 @@ import { Logger } from './logger.js'
...
+ */
+const inferDefinitelyTypedPackage = (specifier: string) => {
+  if (!specifier.startsWith('@')) return `@types/${specifier}`
+  const [scope, pkg] = specifier.split('/')
+  return `@types/${scope.replace('@', '')}__${pkg}`
+}
+
+const detectTypes = async (filePath: string): Promise<string | undefined> => {
+  try {
+    const packageJson = await findUp('package.json', { cwd: filePath })
```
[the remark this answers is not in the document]
Skn0tt: I'm not so sure about this. Have we verified that NFT also includes the `package.json` for CJS modules? Or does it only include `package.json` if that contains `"type": "module"`?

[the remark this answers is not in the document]
eduardoboucas: > Have we verified that NFT also includes the `package.json` for CJS modules?

Yes.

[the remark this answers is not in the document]
Skn0tt: I've tried implementing this, but I don't see how we can identify direct dependencies based on `package.json`. In the test case, here's how the `reason` for `@pt-committee/identidade/package.json` looks:
Screenshot 2023-10-20 at 11 29 03
It references its internals, but not the user code that imports it. So if we wanted to use this to detect the `package.json`, we'd have to first detect the `@pt-committee/identidade/index.js` reason, and then find `@pt-committee/identidade/package.json` from that. That's tricky, and I prefer doing an unneeded `findUp` instead. It's not very costly, only executed in local dev, and makes this easier to reason about.

If you seen an easier way of implementing this that i'm missing, please let me know / push a commit to this!

node/npm_dependencies.ts
```diff
@@ -185,6 +227,13 @@ export const vendorNPMSpecifiers = async ({
     target: 'es2020',
   })
 
+  for (const { barrelName, types } of ops) {
```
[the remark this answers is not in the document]
Skn0tt: done in https://github.com/netlify/edge-bundler/pull/505/commits/59d60b46931e54dbce0f08e8563a11bf0eec79c2

node/npm_dependencies.ts
```diff
@@ -14,6 +15,39 @@ import { Logger } from './logger.js'
...
+
+    const typesPackageJson = join(
+      nodeModulesFolder,
+      inferDefinitelyTypedPackage(packageJsonContents.name),
+      'package.json',
+    )
+    const typesPackageContents = JSON.parse(await fs.readFile(typesPackageJson, 'utf8'))
+    const typesPackageTypes = typesPackageContents.types ?? typesPackageContents.typings
+    if (typesPackageContents) return join(typesPackageJson, '..', typesPackageTypes)
+  } catch {}
```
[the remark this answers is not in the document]
Skn0tt: Did that in https://github.com/netlify/edge-bundler/pull/505/commits/59d60b46931e54dbce0f08e8563a11bf0eec79c2. ESLint forced me to introduce a `safelyDetectTypes` function so that the logic isn't nested too deep.

node/npm_dependencies.ts
```diff
@@ -14,6 +15,39 @@ import { Logger } from './logger.js'
...
+/**
+ * Turns @netlify/functions into @types/netlify__functions.
+ */
+const inferDefinitelyTypedPackage = (specifier: string) => {
+  if (!specifier.startsWith('@')) return `@types/${specifier}`
+  const [scope, pkg] = specifier.split('/')
+  return `@types/${scope.replace('@', '')}__${pkg}`
+}
+
+const detectTypes = async (filePath: string): Promise<string | undefined> => {
```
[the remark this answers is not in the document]
Skn0tt: done in https://github.com/netlify/edge-bundler/pull/505/commits/59d60b46931e54dbce0f08e8563a11bf0eec79c2

node/npm_dependencies.ts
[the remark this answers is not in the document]
eduardoboucas: > That's tricky, and I prefer doing a (not strictly necessary) `findUp` instead. It's not very costly, only executed in local dev, and makes this easier to reason about.

My issue with that is that if you have a module with 10,000 files, you'll be repeating this operation 10,000 times, with no caching whatsoever:

1. Traverse the module tree to find the `package.json`
2. Read the `package.json` from disk
3. Parse it
4. Look for `types`
5. Potentially traverse the module tree again to find `@types/`

It's not costly in the context of the synthetic test case we've put together here, but I don't think we can say the same when we're dealing with real modules.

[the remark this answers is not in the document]
eduardoboucas: On the flip side, you'd do:

1. Look at file: if it's not `package.json`, discard
2. Look up parent of `package.json`, if it's not a direct dependency, discard
3. Look up types (steps 2 to 5 above)

But only once for every node module that is a direct dependency.

node/npm_dependencies.ts
```diff
@@ -185,6 +227,13 @@ export const vendorNPMSpecifiers = async ({
     target: 'es2020',
   })
 
+  for (const { barrelName, types } of ops) {
+    if (!types) continue
+    // we're updating the output instead of adding this to the input,
+    // because esbuild will erase the directive while bundling
+    await prependFile(path.join(temporaryDirectory.path, barrelName), `/// <reference types="${types}" />`)
```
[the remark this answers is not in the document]
Skn0tt: Done in https://github.com/netlify/edge-bundler/pull/505/commits/e4a82a793821ec8cf8427fbd07a81c27aa06482f. Had to change `servePath` to a repo-local directory to test this.

node/npm_dependencies.ts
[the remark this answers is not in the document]
Skn0tt: > Look up parent of package.json, if it's not a direct dependency, discard

In order to properly detect direct dependencies, this means we need to look up the parents of `package.json` within NFT's `reason` object. Added a comment as to why in https://github.com/netlify/edge-bundler/pull/505/commits/444a78f12633c798cec344e27268e5172b1c8657. Assuming that's what you intend?

node/npm_dependencies.ts
```diff
@@ -72,26 +124,39 @@ const getNPMSpecifiers = async (basePath: string, functions: string[], importMap
...
+  const npmSpecifiers: { specifier: string; types?: string }[] = []
   const npmSpecifiersWithExtraneousFiles = new Set<string>()
 
-  reasons.forEach((reason, filePath) => {
-    const packageName = getPackageName(filePath)
+  for (const [packageJsonPath, reason] of reasons.entries()) {
+    // every dependency will have its `package.json` in `reasons` exactly once.
+    // by only looking at this file, we save us from doing duplicate work.
+    const isPackageJson = packageJsonPath.endsWith('package.json')
+    if (!isPackageJson) continue
```
eduardoboucas, line 134: I think this means we won't detect extraneous files? I think we can only discard a file that isn't `package.json` for the purpose of detecting direct dependencies, finding types, etc.

Skn0tt: Good point, yeah! I was assuming we have a test for that, but apparently we don't. Gonna add that later today.

node/npm_dependencies.ts
```diff
@@ -14,6 +15,51 @@ import { Logger } from './logger.js'
...
+ * https://github.com/DefinitelyTyped/DefinitelyTyped#what-about-scoped-packages
+ */
+const inferDefinitelyTypedPackage = (specifier: string) => {
+  if (!specifier.startsWith('@')) return `@types/${specifier}`
+  const [scope, pkg] = specifier.split('/')
+  return `@types/${scope.replace('@', '')}__${pkg}`
+}
+
+/**
+ * Starting from a `package.json` file, this tries detecting a typescript declaration file.
```
eduardoboucas, line 31: nit

```suggestion
 * Starting from a `package.json` file, this tries detecting a TypeScript declaration file.
```

node/npm_dependencies.ts
```diff
@@ -14,6 +15,51 @@ import { Logger } from './logger.js'
...
+ * It first looks at the `types` and `typings` fields in `package.json`.
+ * If it doesn't find them, it falls back to DefinitelyTyped packages (`@types/...`).
+ */
+const detectTypes = async (packageJsonPath: string): Promise<string | undefined> => {
```
eduardoboucas, line 35: nit: I think the name of the method could be a bit more descriptive. "detect" doesn't tell me much about what this returns. I would consider calling this something like `getTypesPath`.

node/npm_dependencies.ts
```diff
@@ -14,6 +15,51 @@ import { Logger } from './logger.js'
 
 const TYPESCRIPT_EXTENSIONS = new Set(['.ts', '.cts', '.mts'])
 
+/**
+ * Returns the name of the `@types/` package used by a given specifier. Most of
+ * the times this is just the specifier itself, but scoped packages suffer a
+ * transformation (e.g. `@netlify/functions` -> `@types/netlify__functions`).
+ * https://github.com/DefinitelyTyped/DefinitelyTyped#what-about-scoped-packages
+ */
+const inferDefinitelyTypedPackage = (specifier: string) => {
```
eduardoboucas, line 24: nit: I would call this `getTypesPackageName`. "Definitely typed" doesn't mean a lot to people who don't know that the repo where the types live have that name, so I think it could be a bit more descriptive.

Also you're calling this "types package" below, so it would make things a bit more consistent.

node/npm_dependencies.ts
```diff
@@ -72,26 +124,39 @@ const getNPMSpecifiers = async (basePath: string, functions: string[], importMap
...
 
     const parents = [...reason.parents]
-    const isDirectDependency = parents.some((parentPath) => !parentPath.startsWith(`node_modules${path.sep}`))
+
+    const isDirectDependency = parents.some((parentPath) => {
+      if (!parentPath.startsWith(`node_modules${path.sep}`)) return true
+      // typically, edge functions have no direct dependency on the `package.json` of a module.
+      // it's the impl files that depend on `package.json`, so we need to check the parents of
+      // the `package.json` file as well to see if the module is a direct dependency.
+      const ownImports = [...(reasons.get(parentPath)?.parents ?? [])]
```
eduardoboucas, line 146: I find this variable name confusing. If I'm reading this right, the variable contains parents, not imports/children?

node/npm_dependencies.ts
```diff
@@ -72,26 +124,39 @@ const getNPMSpecifiers = async (basePath: string, functions: string[], importMap
...
+      return ownImports.some((parentPath) => !parentPath.startsWith(`node_modules${path.sep}`))
+    })
 
     // We're only interested in capturing the specifiers that are first-level
     // dependencies. Because we'll bundle all modules in a subsequent step,
     // any transitive dependencies will be handled then.
     if (isDirectDependency) {
-      const specifier = getPackageName(filePath)
+      const specifier = getPackageName(packageJsonPath)
```
eduardoboucas, line 154: I know that this already existed in the code, but I don't think we need to call `getPackageName` again. We can just use `packageName` from above?

node/npm_dependencies.ts
```diff
@@ -109,27 +174,34 @@ const getNPMSpecifiers = async (basePath: string, functions: string[], importMap
...
 
   return {
-    npmSpecifiers: [...npmSpecifiers],
+    npmSpecifiers,
     npmSpecifiersWithExtraneousFiles: [...npmSpecifiersWithExtraneousFiles],
   }
 }
 
+const prependFile = async (path: string, prefix: string) => {
+  const existingContent = await fs.readFile(path, 'utf8')
```
eduardoboucas, line 186: I really don't like that we are writing a file to disk, then loading it again into memory, and then writing it back to disk.

I think in the future it might be nice to set `write: false` in esbuild, so that we get the file contents directly and write to disk just once.

node/npm_dependencies.ts
Skn0tt: done in https://github.com/netlify/edge-bundler/pull/505/commits/a0a27f0cc7a1beee318a17642450fefe09afcb66

node/npm_dependencies.ts
Skn0tt: renamed in https://github.com/netlify/edge-bundler/pull/505/commits/a0a27f0cc7a1beee318a17642450fefe09afcb66

node/npm_dependencies.ts
Skn0tt: renamed in https://github.com/netlify/edge-bundler/pull/505/commits/a0a27f0cc7a1beee318a17642450fefe09afcb66

node/npm_dependencies.ts
Skn0tt: renamed in https://github.com/netlify/edge-bundler/pull/505/commits/a0a27f0cc7a1beee318a17642450fefe09afcb66

node/npm_dependencies.ts
Skn0tt: fixed in https://github.com/netlify/edge-bundler/pull/505/commits/a0a27f0cc7a1beee318a17642450fefe09afcb66

node/npm_dependencies.ts
```diff
@@ -14,6 +15,51 @@ import { Logger } from './logger.js'
...
+ * If it doesn't find them, it falls back to DefinitelyTyped packages (`@types/...`).
+ */
+const detectTypes = async (packageJsonPath: string): Promise<string | undefined> => {
+  const packageJsonContents = JSON.parse(await fs.readFile(packageJsonPath, 'utf8'))
+  // this only looks at `.types` and `.typings` fields. there might also be data in `exports -> . -> types -> import/default`.
+  // we're ignoring that for now.
+  const packageJsonTypes = packageJsonContents.types ?? packageJsonContents.typings
+  if (typeof packageJsonTypes === 'string') return path.join(packageJsonPath, '..', packageJsonTypes)
+
+  const nodeModulesFolder = await findUp('node_modules', { cwd: packageJsonPath, type: 'directory' })
```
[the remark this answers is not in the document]
Skn0tt: good idea! done in https://github.com/netlify/edge-bundler/pull/505/commits/a0a27f0cc7a1beee318a17642450fefe09afcb66

node/npm_dependencies.ts
Skn0tt: addressed in https://github.com/netlify/edge-bundler/pull/505/commits/e206df474a2ae856b5c7d5bdfb7d1a96eac2f480

node/npm_dependencies.ts
```diff
@@ -109,27 +140,63 @@ const getNPMSpecifiers = async (basePath: string, functions: string[], importMap
...
       })
     }
-  })
+
+    // every dependency will have its `package.json` in `reasons` exactly once.
+    // by only looking at this file, we save us from doing duplicate work.
+    const isPackageJson = filePath.endsWith('package.json')
+    if (!isPackageJson) continue
+
+    const packageJsonPath = filePath
```
eduardoboucas, line 149: nit: This feels a bit odd. Could we not just use `filePath`?

node/npm_dependencies.ts
eduardoboucas: Could we create an issue so we do this as a follow-up?

node/npm_dependencies.ts
```diff
@@ -185,6 +254,15 @@ export const vendorNPMSpecifiers = async ({
     target: 'es2020',
   })
 
+  for (const { barrelName, types } of ops) {
+    if (!types) continue
+    // we're updating the output instead of adding this to the input,
```
eduardoboucas, line 259: It's really annoying that esbuild doesn't make it possible to preserve certain comments in the output (only if they follow the pattern of [legal comments](https://esbuild.github.io/api/#legal-comments)).

Otherwise it'd be super easy for us to add the comment to the barrel file and avoid getting in the business of re-writing files.

😞

node/npm_dependencies.ts
```diff
@@ -14,6 +15,47 @@ import { Logger } from './logger.js'
...
+  const [scope, pkg] = specifier.split('/')
+  return path.join('@types', `${scope.replace('@', '')}__${pkg}`)
+}
+
+/**
+ * Starting from a `package.json` file, this tries detecting a TypeScript declaration file.
+ * It first looks at the `types` and `typings` fields in `package.json`.
+ * If it doesn't find them, it falls back to DefinitelyTyped packages (`@types/...`).
+ */
+const getTypesPath = async (packageJsonPath: string): Promise<string | undefined> => {
```
eduardoboucas, line 35: This is totally a nit, since you shouldn't have to follow any formatting conventions that aren't enforced by our Prettier/ESLint configurations. Here goes anyway, in the spirit of sharing feedback.

I find this function a lot harder to parse than it should be, mostly because the code is super compact. Having some whitespace in between the different "sections" of the logic would help, as would things like separating the condition from the statement in `if` clauses.

Also, if you had the package.json parsing and the `@types/` parsing on two different functions, you could have much shorter variable names (e.g. `packageJsonPath` and `packageJsonContents` could simply be `path` and `contents`), which I think would also help.

Definitely not a blocker, though.

eduardoboucas (dismissed):
🚀

node/npm_dependencies.ts
Skn0tt: Yeah, also checked legal comments :/ The current way is pretty much the easiest, sadly ...

node/npm_dependencies.ts
Skn0tt: Agree, that makes things a lot easier. Thanks for the feedback! I guess I wanted to get this mergeable, so I didn't look too much at readability. Also, it's been a while since I've written PRs with more than 100 lines of new code 😅

Skn0tt: Done in https://github.com/netlify/edge-bundler/pull/505/commits/aa604a8420a24f926db61f6ed56f02d6b61370f7

eduardoboucas (approved):

node/npm_dependencies.ts
Skn0tt: Done in https://github.com/netlify/edge-bundler/pull/505/commits/a8880e05d0913d44139fa5fe9e4f9e2b467396c7

eduardoboucas (approved):

PR #509, opened 2023-10-23
fix: give stable barrel file names

Skn0tt:
Implements https://github.com/netlify/edge-bundler/pull/505#discussion_r1366666275. This also un-flakes the test we added in https://github.com/netlify/edge-bundler/pull/505, since the names no longer relies on the order that files are being read from disk in.

### DIFF
--- a/node/npm_dependencies.ts
+++ b/node/npm_dependencies.ts
@@ -15,17 +15,19 @@ import { Logger } from './logger.js'
 
 const TYPESCRIPT_EXTENSIONS = new Set(['.ts', '.cts', '.mts'])
 
+const slugifyPackageName = (specifier: string) => {
+  if (!specifier.startsWith('@')) return specifier
+  const [scope, pkg] = specifier.split('/')
+  return `${scope.replace('@', '')}__${pkg}`
+}
+
 /**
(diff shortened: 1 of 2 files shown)


node/npm_dependencies.ts
```diff
@@ -233,24 +235,23 @@ export const vendorNPMSpecifiers = async ({
...
+  // To bundle an entire module and all its dependencies, create a entrypoint file
   // where we re-export everything from that specifier. We do this for every
   // specifier, and each of these files will become entry points to esbuild.
   const ops = await Promise.all(
-    npmSpecifiers.map(async ({ specifier, types }, index) => {
+    npmSpecifiers.map(async ({ specifier, types }) => {
       const code = `import * as mod from "${specifier}"; export default mod.default; export * from "${specifier}";`
-      const barrelName = `barrel-${index}.js`
-      const filePath = path.join(temporaryDirectory.path, barrelName)
+      const filePath = path.join(temporaryDirectory.path, `bundled-${slugifyPackageName(specifier)}.js`)
```
eduardoboucas, line 244: I'm curious why you went with the `bundled-` prefix here, rather than just using the slugified package name?

eduardoboucas (approved):

PR #517, opened 2023-10-30
fix: prefer ESM if available

Skn0tt:
Discovered this via https://answers.netlify.com/t/support-for-npm-modules-in-edge-functions/101645/3?u=eduardoboucas.

The module `@optmizily/optimizely-sdk` ships both ESM and CJS files, but doesn't have a `type: "module"` declaration. Here's an excerpt from its `package.json`: 

```json
{
  "name": "@optimizely/optimizely-sdk",
  "module": "dist/optimizely.browser.es.js",
  "main": "dist/optimizely.node.min.js",
  "browser": "dist/optimizely.browser.min.js",
  "react-native": "dist/optimizely.react_native.min.js",
  "typings": "dist/index.browser.d.ts",
```

As documented in https://esbuild.github.io/api/#main-fields, Node.js will consume `main` in this case. By default, ESBuild mimics this behaviour. This results in ESBuild trying to transpile CJS to ESM, and in the example of the forums post above, this broke named imports.

We can fix this by preferring the `module` or `browser` distributions, if available.


mraerino (commented):
is this a non-breaking change? do we need a feature flag?

Skn0tt:
I'm pretty sure it's non-breaking, and it's for an experimental feature. Don't think we need a featureflag.

mraerino (approved):

PR #519, opened 2023-11-02
fix: parse TSX files for module detection, define NODE_ENV, polyfill missing Node.js globals

Skn0tt:
This PR addresses three issues:

1. we weren't correctly detecting NPM modules in `.tsx` files, because we didn't transpile them
2. modules that ship dev builds, like `react`, are currently failing to bundle because they rely on `process.env.NODE_ENV` to detect which build to load - but `process.env` is not defined. We're polyfilling `process`, and a couple other Node.js globals.
3. when bundling `react` for prod, there's no reason to include the development build. we're defining `NODE_ENV=production` during prod builds, so esbuild can tree-shake the development build.


node/npm_dependencies.ts
```diff
@@ -267,6 +268,10 @@ export const vendorNPMSpecifiers = async ({
     splitting: true,
     target: 'es2020',
     write: false,
+    define: {
+      // unsure if this is a wise setting, but it seems to work for this case. should we set "development" for local CLI?
+      'process.env.NODE_ENV': '"production"',
```
Skn0tt, line 273: Because we're bundling a Node module to become a Deno module, no Node globals are available. Deno's native NPM support supplies these by default, we'll have to work around this. A couple thoughts:

1. if we can find a way to tell Deno to treat this file as a Node module, that'd be ideal
1. In local dev, maybe a polyfill that populates globalThis.process with Deno's node:process would work better here
1. In builds, I think it's good to define this during bundling, so we don't bundle too much code in. But we should probably not hard-code it.

node/npm_dependencies.ts
```diff
@@ -110,6 +110,7 @@ const getNPMSpecifiers = async (
     readFile: async (filePath: string) => {
       // If this is a TypeScript file, we need to compile in before we can
       // parse it.
+      // mabye we should make this `path.extname(filepath).includes("ts")`?
       if (TYPESCRIPT_EXTENSIONS.has(path.extname(filePath))) {
```
Skn0tt, line 114: mabye we should make this `path.extname(filepath).includes("ts")`?

eduardoboucas: Why?

eduardoboucas:
You have a failing test.

node/npm_dependencies.ts
Skn0tt: it wouldn't require us to update the set everytime we learn about a new typescript extension ;D happy to keep it how it is, though

Skn0tt:
Updated the code to only set `process.env.NODE_ENV` for prod bundles, but leaving it unset for development bundles. Addressing the polyfills now.

node/npm_dependencies.ts
```diff
@@ -244,12 +246,26 @@ export const vendorNPMSpecifiers = async ({
...
+    `
+    import process from "node:process";
+    import { setImmediate, clearImmediate } from "node:timers";
+    import { Buffer } from "node:buffer";
+    globalThis.process = process;
+    globalThis.Buffer = Buffer;
+    globalThis.setImmediate = setImmediate;
+    globalThis.clearImmediate = clearImmediate;
+    globalThis.global = globalThis;
+    `,
```
Skn0tt, line 260: I came up with this list by going through https://nodejs.org/api/globals.html, and checking which of those don't exist in Deno. I think I caught all of them.
The only ones missing are PerformanceObserver, PerformanceObserverEntryList and PerformanceResourceTiming, but something tells me that the above are way more important, and we could get away without those.

Skn0tt:
https://github.com/netlify/edge-bundler/pull/519/commits/66161b87cf005c2778957315ad3fb3d320f895d2 adds polyfills for Node.js Globals that don't exist within Deno. This will fix any packages relying on those, but it has the downside of polluting the global namespace - so those globals will also be accessible in user code, once any NPM library is imported. I'm not sure if we want that. The alternative would be some sort of compile-time rewrite. Now that I think about it, maybe we could use an ESBuild Banner for this?

Skn0tt:
Yes, the Banner works way better :)

node/bundler.test.ts
```diff
@@ -536,3 +537,33 @@ test('Loads JSON modules', async () => {
...
+
+  await bundle([sourceDirectory], distPath, declarations, {
+    basePath,
+    vendorDirectory: vendorDirectory.path,
+  })
+
+  const manifestFile = await readFile(resolve(distPath, 'manifest.json'), 'utf8')
+  const manifest = JSON.parse(manifestFile)
+  const bundlePath = join(distPath, manifest.bundles[0].asset)
+  process.env.FOO = 'bar'
```
eduardoboucas, line 560: Can we clean up after this?

node/npm_dependencies.ts
```diff
@@ -81,6 +81,10 @@ const safelyDetectTypes = async (packageJsonPath: string): Promise<string | unde
 // Workaround for https://github.com/evanw/esbuild/issues/1921.
 const banner = {
   js: `
+  import process from "node:process";
+  import { setImmediate, clearImmediate } from "node:timers";
```
eduardoboucas, line 85: nit: Can we make the formatting match the imports we already had? No preference on which convention we pick, but let's try to be consistent.

node/npm_dependencies.ts
```diff
@@ -209,6 +213,7 @@ interface VendorNPMSpecifiersOptions {
   importMap: ImportMap
   logger: Logger
   referenceTypes: boolean
+  bundleProd: boolean
```
eduardoboucas, line 216: We already have a boolean parameter (`referenceTypes`) that we toggle based on whether we're bundling for production or local development. Rather than adding a new one, can we have a single `production: boolean` or `environment: "production" | "dev"` that we use for both?

node/bundler.test.ts
Skn0tt: done in https://github.com/netlify/edge-bundler/pull/519/commits/5a9c7f3ec43f3e81c2fb3c11ce94edbe4f0fd3fc

node/npm_dependencies.ts
Skn0tt: done in https://github.com/netlify/edge-bundler/pull/519/commits/0e45e3b23df5aad5fd0d3d9fabf8f5cd830497f8

node/npm_dependencies.ts
Skn0tt: done in https://github.com/netlify/edge-bundler/pull/519/commits/b8394e8803849819f3e56877754f8b658a9a5c7d

test/fixtures/tsx/functions/func1.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
```
eduardoboucas, line 1: Are we getting anything from actually including React in a fixture and committing its source to version control? It's fine to do it if we think it will make the test more accurate, but it feels that in this case we only want to check if a global is defined and not even using the module itself (which means it's probably getting tree-shaken)?

eduardoboucas (approved):

test/fixtures/tsx/functions/func1.tsx
Skn0tt: If you take a look at the sources, you'll see that it's two placeholders and there's no real React in there. Happy to rename to a more telling name, if that makes it less confusing.

eduardoboucas: Oh, I see! GitHub hides the source of those files by default and I had assumed they were the real thing. Sorry, ignore me!

PR #520, opened 2023-11-02
fix: pass cache object to nft to save some work

Skn0tt:
Part of https://linear.app/netlify/issue/COM-72/edge-functions-local-dev-dont-delete-files-on-every-build. 

Passing a cache object into NFT allows to save some work during local dev.

Next step is to use an [ESBuild Context](https://esbuild.github.io/api/#build) to save some work over there, but that requires a bigger refactoring - not sure if it's worth it.


Skn0tt:
Today I learned that NFT's cache doesn't check if the file got updated before reading from cache: https://github.com/vercel/nft/blob/99692f6c092bc174b2ffc2f21efb7a8afc42af57/src/fs.ts#L47-L48

That means it's entirely unusable for a long-running dev server :/

PR #521, opened 2023-11-04
feat: add `rootPath` for monorepo setups

eduardoboucas:

Support for npm modules is currently broken for monorepos, since we're setting `basePath` to the package root, but this means we won't be able to find any npm modules in the repository root, i.e. outside of `basePath`.

To address that, we introduce a new `rootPath` that should always be set to the repository root. In non-monorepo setups, `basePath` and `rootPath` will be the same.


ascorbic (approved):
Nice!

ascorbic (approved):

PR #528, opened 2023-11-08
feat: add `ModuleGraph` type

eduardoboucas:

The return value of `serve` includes a `graph` property that contains the module graph for the functions. This is currently untyped, which makes it harder for consumers (like the CLI) to process this module safely.

This PR adds a `ModuleGraph` type, extracted from https://deno.land/x/deno_graph@0.59.2/types.d.ts, which we're now vendoring.

I've also added a new `vendor` npm script for recreating the `deno/vendor` directory with all the modules we're referencing.


README.md
```diff
@@ -9,32 +9,42 @@ Intelligently prepare Netlify Edge Functions for deployment.
...
+   ```
+
+## Vendored modules
+
+To avoid pulling in additional dependencies at runtime, this package vendors some Deno modules in the `deno/vendor`
+directory.
+
+You can recreate this directory by running `npm run vendor`.
+
+> [!WARNING]
```
Skn0tt, line 43: nice, thanks for documenting this!

Skn0tt (approved):

eduardoboucas:
Ah, I had to push https://github.com/netlify/edge-bundler/pull/528/commits/adf717e15bc2c8ef2bb3b666a0deca2ae12d4ca4. Can I get another ✅, please?

Skn0tt (approved):

PR #535, opened 2023-11-15
fix: prevent global namespace clash for `Buffer`

Skn0tt:
When bundling an NPM package that does `import { Buffer } from "node:buffer"`, like `@netlify/blobs`, then edge-bundler was outputting both the unchanged import from said package, plus another import via the Banner added in #519. Here's an excerpt for how the output looks:

```js

  ...
  import {Buffer} from "node:buffer";
  import {createRequire as ___nfyCreateRequire} from "node:module";
  import {fileURLToPath as ___nfyFileURLToPath} from "node:url";
  import {dirname as ___nfyPathDirname} from "node:path";
  let __filename=___nfyFileURLToPath(import.meta.url);
  let __dirname=___nfyPathDirname(___nfyFileURLToPath(import.meta.url));
  let require=___nfyCreateRequire(import.meta.url);
  

// test/fixtures/imports_npm_module/node_modules/child-1/index.js
import { Buffer } from "buffer";
...
```

Those two import lines clash at runtime. This PR fixes this, by populating the global namespace via writing onto `globalThis`. This solves it because it's shadowed by imports, but has the drawback that it also affects `globalThis` in user-code (see changed test). I'll see if I can find a workaround for this.

### DIFF
--- a/node/npm_dependencies.ts
+++ b/node/npm_dependencies.ts
@@ -81,16 +81,19 @@ const safelyDetectTypes = async (packageJsonPath: string): Promise<string | unde
 // Workaround for https://github.com/evanw/esbuild/issues/1921.
 const banner = {
   js: `
-  import process from "node:process";
-  import {setImmediate, clearImmediate} from "node:timers";
-  import {Buffer} from "node:buffer";
-
+  import __nfyProcess from "node:process";
+  import {setImmediate as __nfySetImmediate, clearImmediate as __nfyClearImmediate} from "node:timers";
+  import {Buffer as __nfyBuffer} from "node:buffer";
   import {createRequire as ___nfyCreateRequire} from "node:module";
   import {fileURLToPath as ___nfyFileURLToPath} from "node:url";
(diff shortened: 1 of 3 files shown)


Skn0tt:
> the drawback that it also affects globalThis in user-code (see changed test). I'll see if I can find a workaround for this.

I tried moving the Buffer import into a polyfill file, but couldn't get that to work :/ @eduardoboucas if you have any idea, i'm all ears!

eduardoboucas:
> I tried moving the Buffer import into a polyfill file, but couldn't get that to work :/ @eduardoboucas if you have any idea, i'm all ears!

If I'm reading this right, we're not only affecting user code, but doing it only in certain conditions — i.e. `Buffer` will be available in the global scope only if you use npm modules, because otherwise we won't add the esbuild step?

eduardoboucas (approved):
The issues we've identified are definitely not ideal, but I think shipping this is much better in the short-term than breaking functions.

PR #548, opened 2023-12-07
feat!: provide import maps when starting the isolate, not server

Skn0tt:
During local dev, the import map paths can change after startup. For example when a framework generates an import map after the dev server was already initiated. So we can't know all the paths up-front!

Currently, edge-bundler needs us pass all of them at startup. But it doesn't need them until invocation time.

This PR moves the import map parameter from startup time to invocation time, so we don't have to work around this in the CLI.

### DIFF
--- a/node/server/server.ts
+++ b/node/server/server.ts
@@ -183,7 +183,6 @@ interface ServeOptions {
   distImportMapPath?: string
   featureFlags?: FeatureFlags
   inspectSettings?: InspectSettings
-  importMapPaths?: string[] | (() => Promise<string[]>)
   onAfterDownload?: OnAfterDownloadHook
   onBeforeDownload?: OnBeforeDownloadHook
   formatExportTypeError?: FormatFunction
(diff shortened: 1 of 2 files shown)


node/server/server.ts
```diff
@@ -181,7 +183,7 @@ interface ServeOptions {
   distImportMapPath?: string
   featureFlags?: FeatureFlags
   inspectSettings?: InspectSettings
-  importMapPaths?: string[]
+  importMapPaths?: string[] | (() => Promise<string[]>)
```
eduardoboucas, line 186: Can you explain when we might want `importMapPaths` to be a function? I don't see us calling it anywhere?

Skn0tt: Oh I think that's a remnant!

Skn0tt: removed in https://github.com/netlify/edge-bundler/pull/548/commits/556a32d4e0432217d9ab292d225a88e386533f29

eduardoboucas (approved):

PR #560, opened 2024-01-10
feat: transform negative lookaheads

eduardoboucas:

When processing a declaration pattern, transforms negative lookaheads into a set of exclusion patterns.

Supersedes #451. Part of FRA-195.

### DIFF
--- a/node/declaration.ts
+++ b/node/declaration.ts
@@ -1,4 +1,4 @@
-import regexpAST from 'regexp-tree'
+import regexpAST, { NodePath } from 'regexp-tree'
 
 import { FunctionConfig, HTTPMethod, Path } from './config.js'
 import { FeatureFlags } from './feature_flags.js'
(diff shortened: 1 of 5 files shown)


node/declaration.ts
```diff
@@ -116,20 +116,73 @@ const createDeclarationsFromFunctionConfigs = (
...
+ * Validates and normalizes a pattern so that it's a valid regular expression
+ * in Go, which is the engine used by our edge nodes.
+ *
+ * @param pattern Original regular expression
+ * @param extractedExclusionPatterns If set, negative lookaheads (which are
+ * not supported in Go) are transformed into a list of exclusion patterns to
+ * be stored in this array
+ * @returns Normalized regular expression
+ */
+export const parsePattern = (pattern: string, extractedExclusionPatterns?: string[]) => {
```
Skn0tt, line 129: nit: Could we return `extractedExclusionPatterns` as an array, instead of mutating the parameter?

Skn0tt (approved):
The approach looks good and I like that we're keeping this behind a flag. Let's see how this performs!

node/declaration.ts
eduardoboucas: That would be the cleanest way, for sure. But `parsePattern` is used in a few different contexts and I wanted to change the API surface as little as possible to contain the change behind the flag. We can revisit this in the future.

node/declaration.ts
```diff
@@ -146,6 +199,79 @@ export const parsePattern = (pattern: string) => {
     },
   })
 
+  // The `extractedExclusionPatterns` property works as a shut-off valve: if
+  // it's not supplied, don't even traverse the AST again to further process
+  // lookaheads.
+  if (extractedExclusionPatterns) {
+    const exclusionPatterns = [...lookaheads].map((lookahead) => getExclusionPatternFromLookahead(regexp, lookahead))
+
+    extractedExclusionPatterns.push(...exclusionPatterns)
```
mraerino, line 208: have you thought about introducing a logging-only mechanism that can give you data on real-world examples so you can spot check if this code is correct?

PR #563, opened 2024-01-15
feat: support PCRE regexp engine

eduardoboucas:

As we add a PCRE-compatible regular expression engine to our edge nodes, we must stop transforming regular expressions in Edge Bundler to make them compatible with the RE2 engine. We're doing this behind a feature flag.

Part of https://linear.app/netlify/issue/FRA-195/devise-strategy-to-handle-nextjs-routing-matchers.


Skn0tt (approved):

orinokai (approved):
lgtm!

